PR #26, opened 2024-03-21
Add the dataclasses exposing the ER model to the rest of the app.

yoannLafore:
This pull request aims to provide the necessary dataclasses to expose the following notions of the ER model to the higher levels of the app:

User: 
 - CurrentUser: Which is the current user of the app with all his informations.
 - ForeignUser: Which is another user of the app that the current user has no link with.

Post:
 - Which is posted by a user and has a list of comments represented by `loadedComments` because not every comments will be loaded at the same time.

Comment:
 - Which is posted by a user under a post or a comment and has a list of anwsers represented by `loadedAnwsers` for similar reasons.

Location:
 - Simple location on the earth to locate a post.


gruvw:
Ok so I took the time to review the PR, and it looks great except for the following remarks:

- I don't think that the current user using the application should extend the `User` class. The two classes are really not related, the way we will make use of them. The user using the application essentially needs to have a user id and metadata like their email (that we get from Google Account). When we want to display their profile, we will pull the data from the database using their ID and parse it into the `User` class.
- The name of the classes that you chose will definitely conflict with existing class names (example `User` already used by Firebase). They must be renamed with something either a little bit more descriptive (in order to avoid conflict) or prefixed with a common prefix (like `Proxi` or `Data` ...).

Other than that, the PR looks ok :+1:

yoannLafore:
Thanks for your remarks !
I have changed my approach to represent the user's data instead of the user directly (which will be handled with the firebase logic).
So I have removed the previous `User` classes and added the following:
`CurrentUserData`: Which contains the data of the user currently using the app (those are the data that will be fetched when the user navigates to his profile).

`ForeignUserData`: Which contains the public data of a user (those will be queried to display the reduced profile of a post owner above the post for example).

gruvw:
Still a few things:

- Do we really want a username AND a display name ? Can't we just have the `@username` be the display name of the participants ?
- I think you kinda mixed up the Data classes that are directly related to the Firestore database collections and the Data classes that are related to the UI: for example you have the `loadedComments` attribute which implies that they are not all loaded by default, meaning it does not correspond 1 to 1 to the Firestore database.
- For the previous reason I think it would make a lot of sense if you create the Firestore collections on the Firestore console directly with this PR and add a constructor with the corresponding data class to ensure the tha Data classes correspond to the actual implementation (just wait for #25 to be merged and pull the changes inside your branch afterward, there should be no conflict). You can take an example on the following SO answer https://stackoverflow.com/a/52316762/14580601 and add a named constructor `.fromDb` and a method `toDb()` to every data class.
- Along with that you will be able to verify correct integration of the Firestore database with the Data classes by writing unit tests using the Firestore mocks.

I think the PR will make more sense that way so we have a full basis for the data already connected to the Firestore schema and ready to be used.

Issue #17, opened 2024-03-19
Firestore database dataclasses + communication


JoachimFavre:
Add the dataclasses to get a one-to-one mapping the following firestore documents: `Post` and `User`.
Create the repositories to perform the querying of those documents.

PR #43, opened 2024-03-27
Add repositories for user and posts

yoannLafore:
Hello,

This PR implements #17.
This PR aims to provide the necessary components to allow for the querying of users and posts.
I know that this PR is quite big (most of it are tests) and in the future I'll try to subdivide such tasks into multiple PRs.

So, for the users, this PR provides a representation of the user as it is stored on firestore. In particular the PR provides a repository that implements:
 - Set and get the current user data : `setCurrentUser`, `getCurrentUser`
 - Get if a user exists or not : `doesUserExists`
 - Get an arbitrary user : `getUser`

Regarding the posts, the PR once again provides a representation of the posts stored on firestore. In particular, the PR provides a repository that implements:
 - Get the posts near the user : `getNearPosts`
 - Get a post by its id : `getPost`
 - Add a post at the current position of the user : `addPost`
 - Delete a post : `deletePost`

I thank you for reviewing my PR and look forward for you feedback and suggestions :)


gruvw:
Hello, the PR looks good but needs some refactoring and provider integration.

I plan to implement those tomorrow :+1:

lib/models/database/post_repository.dart
```diff
@@ -0,0 +1,278 @@
...
+    });
+  }
+
+  Future<List<PostFirestore>> _getPostsNearLocation(
+    GeoFirePoint point,
+    double radius,
+  ) async {
+    final posts = await _geoFire
+        .collection(collectionRef: _collectionRef)
+        .within(
```
JoachimFavre, line 252: I am not sure at all of this comment. In [the documentation](https://pub.dev/packages/geoflutterfire2), they say that `within` may cause "memory leak, because underlying stream listeners are not cancelled when you cancel StreamSubscription from within method", and that `withinAsSingleStreamSubscription` is much safer. I was wondering if you thought about this and `within` was a concious choice? 
I really struggle to understand the (too small) documentation of this package, but, since the stream is cancelled imediately after on the `first`, don't we have a memory leak?

lib/models/database/user_repository.dart
```diff
@@ -0,0 +1,147 @@
...
+    return UserFirestore.fromDb(docSnap);
+  }
+
+  Future<void> setCurrentUser(UserFirestoreData userData) async {
+    final uid = _getUid();
+
+    await _collectionRef.doc(uid).set(userData.toDbData());
+  }
+
+  Future<bool> doesUserExists(UserFirestoreId uid) async {
```
JoachimFavre, line 133: Small typo, this should be `doesUserExist` (without an s at the end)

test/models/database/post_repository_test.dart
```diff
@@ -0,0 +1,335 @@
...
+import "package:fake_cloud_firestore/fake_cloud_firestore.dart";
+import "package:flutter_test/flutter_test.dart";
+import "package:geoflutterfire2/geoflutterfire2.dart";
+import "package:mockito/mockito.dart";
+import "package:proxima/models/database/post_repository.dart";
+import "package:proxima/services/geolocation_service.dart";
+
+import "mock_post_data.dart";
+
+class MockGeoLocationService extends Mock implements GeoLocationService {
```
JoachimFavre, line 11: I find a bit weird that is defined here and not in test/services/mock_geo_location_service? However, it is indeed only used in here; so if you don't think any other test will ever need it, I guess it may be fine leaving it here.

test/models/database/post_repository_test.dart
```diff
@@ -0,0 +1,335 @@
...
+
+    test("delete post correctly", () async {
+      await setPostFirestore(post);
+
+      await postRepository.deletePost(post.id);
+      final actualPost = await firestore
+          .collection(PostRepository.collectionName)
+          .doc(post.id)
+          .get();
+      expect(actualPost.exists, false);
```
JoachimFavre, line 186: I would personally find the test more logical if it verifies the post exists before deleting it. Currently, the `setPostFirestore` could not update the database and we would not be able to notice. Adding this verification, we would indeed see that a post which exists, no longer exists after being deleted.

test/models/database/user_repository_test.dart
```diff
@@ -0,0 +1,162 @@
+import "package:cloud_firestore/cloud_firestore.dart";
+import "package:fake_cloud_firestore/fake_cloud_firestore.dart";
+import "package:firebase_auth_mocks/firebase_auth_mocks.dart";
+import "package:flutter_test/flutter_test.dart";
+import "package:proxima/models/database/user_repository.dart";
+
+void main() {
+  group("User repository testing", () {
```
JoachimFavre, line 8: You don't have a test for hashing in this one, as you have for the `PostRepository` tests. I don't think this is strictly necessary to implement (I'm not sure I would have added a test for hashes in the first place, even though more testing is better than less testing), but I just wanted to point it out.

test/services/geolocation_service_test.dart
```diff
@@ -0,0 +1,166 @@
...
+
+    test("getCurrentPosition returns GeoPoint", () async {
+      const latitude = 37.4219983;
+      const longitude = -122.084;
+      const expectedGeoPoint = GeoPoint(latitude, longitude);
+
+      when(mockGeolocator.isLocationServiceEnabled())
+          .thenAnswer((_) async => true);
+      when(mockGeolocator.checkPermission())
+          .thenAnswer((_) async => LocationPermission.always);
```
JoachimFavre, line 78: Isn't this a repetition with lines 26-39?
I would personally think that the `when` should be prefered over the `noSuchMethod` since it is more readable (if it is possible to do once and for all in the setup). Note that, changing `MockGeolocatorPlatform` would also mean changing `post_repository_test.dart::MockGeoLocationService`.

JoachimFavre (commented):
Wow, that was a lot of code to read. This is very nice, it's a lot of non-trivial code we really needed to move forward, thank you so much! :)
There are so many tests too, that's great!

You can find my comments below. I also have some more general comments. The first one is that I feel some lack of higher-level documentation. For instance, I would have enjoyed some more documentation specifying what `PostFirestore` should be used for, or the nuance between `UserRepository::getCurrentUser()` and `UserRepository::getUser()` without looking at the code (don't get me wrong, those functions are named perfectly well, and they make a lot of sense once we know what they do, but a small line of documentation would have allowed me to understand faster).

The second thing is about code structure. I'll let @gruvw  develop the refactoring and providers, he knows more than me on this subject and seemed to want to do it tomorrow. I however find that your files `post_repository.dart` and `user_repository.dart` are quite cluttered, with many classes. I'm for instance not sure I understand why you put `PostFirestore` and `PostRepositroy` in the same file. I however like the fact that the `PostFirestore` content is split in a total of three classes, and that it allows to simplify a lot the interface of `PostRepository` (though, adding this reason somewhere would have been useful for me, even if I'm not sure where you could put this meaningfully).

Aderfish:
Good job! The code looks good to me and well tested. I also think we could refactor a bit and put the data classes and repository classes in different files.

lib/models/database/post_repository.dart
```diff
@@ -0,0 +1,278 @@
...
+  const PostFirestore({
+    required this.id,
+    required this.location,
+    required this.data,
+  });
+
+  /// Parses the data from a firestore document
+  factory PostFirestore.fromDb(DocumentSnapshot docSnap) {
+    if (!docSnap.exists) {
+      throw Exception("Post document does not exist");
```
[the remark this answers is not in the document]
yoannLafore: I agree on the difference between `Exception` and `Error`. However, I don't think that we should throw an error when the document doesn't exist because this something that could be expected (for instance, someone gets a post and at the same time someone else just deleted it). Thus, higher levels might want to catch it to display an error message. 

`FormatException` seems to be suitable for our usage (while not ideal as you said) so I will use them for parse exceptions, thanks for the suggestion !
The commit 1558e9f32d640d6b38f931739625caf8580c6613 add those.

lib/services/geolocation_service.dart
```diff
@@ -0,0 +1,49 @@
+import "package:cloud_firestore/cloud_firestore.dart";
+import "package:geolocator/geolocator.dart";
+
+class GeoLocationService {
+  final GeolocatorPlatform _geoLocator;
+
+  // Settings to get the most accurate location
```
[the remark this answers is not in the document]
yoannLafore: Yes you are right, it makes more sense to use `LocationAccuracy.best` for our setting.
The commit 2af73d94327d29b8c97af7cafeb13524db45e11a fix that.

lib/services/geolocation_service.dart
```diff
@@ -0,0 +1,49 @@
...
+    if (permission == LocationPermission.denied) {
+      permission = await _geoLocator.requestPermission();
+      if (permission == LocationPermission.denied) {
+        // Permissions are denied after requesting
+        throw Exception("Location permissions are denied");
+      }
+    }
+
+    if (permission == LocationPermission.deniedForever) {
+      // Permissions are denied forever, handle appropriately.
```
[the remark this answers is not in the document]
yoannLafore: Correct, I add this the product backlog and update the comment. Thanks !
The commit 81c8f5d28ec15c4f23d3f023fefcbb6dd21dbde4 fix that.

lib/models/database/post_repository.dart
```diff
@@ -0,0 +1,278 @@
+import "package:cloud_firestore/cloud_firestore.dart";
+import "package:flutter/foundation.dart";
+import "package:geoflutterfire2/geoflutterfire2.dart";
+import "package:proxima/models/database/user_repository.dart";
+import "package:proxima/services/geolocation_service.dart";
+
+/// The id are strong typed to avoid misuse
+typedef PostFirestoreId = String;
```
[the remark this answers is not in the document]
yoannLafore: Yes you're correct. I will add those class. 
Unfortunately, dart doesn't allow to create class inside another class thus I will create those class outside.
The commit 48361587dfb9be0ddb5f9211aa208327dadf2736 fix this.

lib/services/geolocation_service.dart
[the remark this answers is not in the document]
JoachimFavre: Thanks!
What about the comment "Settings to get the most accurate location", though? I have to admit I still consider it is not correct. In my opinion, explaining your choice of `best` over `high` and `bestForNavigation` in it would be better.

lib/services/geolocation_service.dart
```diff
@@ -0,0 +1,49 @@
...
+  // Settings to get the most accurate location
+  final LocationSettings locationSettings = AndroidSettings(
+    accuracy: LocationAccuracy.high,
+  );
+
+  GeoLocationService({
+    required GeolocatorPlatform geoLocator,
+  }) : _geoLocator = geoLocator;
+
+  Future<GeoPoint> getCurrentPosition() async {
```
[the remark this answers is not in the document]
yoannLafore: Yes, I have added a reference at the beginning but then have removed it after modifying the code.
But I will add it back.
The commit b3068d9f9472a40e020f5b28e5b0aa770d76e8f5 add the reference.

lib/models/database/post_repository.dart
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  @override
+  int get hashCode {
+    return Object.hash(id, location, data);
+  }
+}
+
+@immutable
+class PostFirestoreData {
```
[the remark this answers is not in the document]
yoannLafore: Yes you're totally right, it's important to keep a consistent notation.
I'll use the format similar to `PostDataFirestore`.
The commit 9ced0490901b1d0efbfa70d7ef6e7f5505c9a376 apply this naming convention.

lib/models/database/post_repository.dart
```diff
@@ -0,0 +1,278 @@
...
+    GeoFirePoint point,
+    double radius,
+  ) async {
+    final posts = await _geoFire
+        .collection(collectionRef: _collectionRef)
+        .within(
+          center: point,
+          radius: radius,
+          field: PostFirestore.locationField,
+          strictMode: true,
```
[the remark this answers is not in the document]
yoannLafore: After digging through the code library, it appears that strict mode simply applies a filter on the query but is not related to approximation of the Earth topology.
However, because we already apply such a filter in our function, we can effectively set strict mode to false.


Code from the library:
```dart
      final filteredList = strictMode
          ? mappedList
              .where((DistanceDocSnapshot doc) =>
                      doc.distance! <=
                      radius * 1.02 // buffer for edge distances;
                  )
              .toList()
          : mappedList.toList();
```

lib/models/database/post_repository.dart
```diff
@@ -0,0 +1,278 @@
...
+        )
+        .first;
+
+    return posts.map((docSnap) => PostFirestore.fromDb(docSnap)).where((post) {
+      // We need to filter the posts because the query is not exact
```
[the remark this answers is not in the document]
yoannLafore: Once again, because of the poor documentation, I digged through the code to find out that with strict mode, the `within` function would still add a 2% margin. So, I decided to add real strict filter in out function.

Code from the library:
```dart
      final filteredList = strictMode
          ? mappedList
              .where((DistanceDocSnapshot doc) =>
                      doc.distance! <=
                      // Here they add a 2% margin
                      radius * 1.02 // buffer for edge distances;
                  )
              .toList()
          : mappedList.toList();
```

lib/models/database/post_repository.dart
yoannLafore: You're right, I have stubble accross this issue and tried to found an alternative but somehow I completely missed the `withinAsSingleStreamSubscription` method (I don't really know how).

So thanks for pointing it out and I fix it in commit 0b1ed553ca519807cfb50389256f908ee725c2f5.

lib/models/database/user_repository.dart
yoannLafore: Yep, corrected in 47a4e5bd8a4449a5739b4af036d8021f70b5ba59 :+1: .

test/models/database/post_repository_test.dart
yoannLafore: Yes you are right, we might need it elsewhere, I'll follow your suggestion.
Fix in 06c590e82f11d0092f47bb084db5728c4e3139ae.

test/models/database/post_repository_test.dart
yoannLafore: Yes, this should make the test more robust, fix in 4a6b6bcb4694b2e67dbebafed8d3199b6c521131.

test/models/database/user_repository_test.dart
yoannLafore: Yes, I think I forgot, I'll add it for consistency. 
Fix in 57d2ca7a0df7e3f6b09b9568cde48d7c289e6b66.

test/services/geolocation_service_test.dart
yoannLafore: It is indeed necessary, the overrides with the `noSuchMethod` are a boiler plate needed in order mock the result with `when`. 
If those are removed, the function won't be mocked properly and a bad state exception will be thrown by the mocking service.

Also, I am not entirely sure of my understanding of `noSuchMethod` but I think that if they are called without having mocking it with `when` priorly, they will throw an exception. This what I have noticed while coding the tests.

So I think that we need to let those here in order for the mocking to work properly.

gruvw:
So we refactored the data classes with @yoannLafore, we also split some concerns to only take care of the database part and leave logic to the VM.

Tests need some adaptations in order to work with the new files / design.

yoannLafore:
Thanks @gruvw for helping refactoring the code.
I have adapted the tests so that they are consistent with the new refactored version of the code.
I also have added documentation to the methods of the repository, please tell me if you think that further documentation is needed.

I hope those changes will make  the code clearer and I look forward to your feedback :)

Aderfish (approved):

gruvw:
Good job on refactoring the tests, you still have quite a few test data duplication: for example a project wide search for "username_8456" gives 6 results.

Can you still try to always reuse the `testingUserData` from the `test/services/test_data/firestore_user_mock.dart` file in all your tests?

Other than that, I think it is good to merge for me.

yoannLafore:
You are correct, thanks for pointing it out :+1: 
I correct this in commit 423fd776fd3ac94c7e64ed4f2281b0f01b481428.

gruvw (approved):

lib/models/database/post_repository.dart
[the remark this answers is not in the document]
JoachimFavre: Wow, you even took a look at the library code. Good job! :)

test/services/geolocation_service_test.dart
JoachimFavre: Oh, we learn every days! Thank you, and sorry for my mistake :)

lib/services/database/user_repository_service.dart
```diff
@@ -0,0 +1,43 @@
...
+  /// This method will retrieve the user with id [uid] from the database
+  Future<UserFirestore> getUser(UserIdFirestore uid) async {
+    final docSnap = await _collectionRef.doc(uid.value).get();
+
+    return UserFirestore.fromDb(docSnap);
+  }
+
+  /// This method will set the user with id [uid] to have the data [userData]
+  /// If the user does not exist yet, it will be created
+  Future<void> setUser(UserIdFirestore uid, UserData userData) async {
```
JoachimFavre, line 27: How do we get a `UserIdFirestore` for the logged in user? I am guessing this can be extracted, but I can't seem to find where

JoachimFavre (changes requested):
I can't believe how much you @yoannLafore and @gruvw worked on this PR, you did an amazing job; what you have implemented will be very useful.
I just have a small comment, you can find below. This PR will be merged very soon! :))

lib/services/database/user_repository_service.dart
yoannLafore: Thanks for the comment ! This is effectively a point that I have not really well explained.
The firestore collection will be constructed such that the id of the user document is the same as its id from Firebase Auth.
So in order to get the `UserIdFirestore`, we have to get the uid of the current user through Firebase Auth.
Here is a snippet of how we could do it:

```dart
    final firebaszeAuth = FirebaseAuth.instance;
    final userId = firebaszeAuth.currentUser?.uid;
    if (userId == null) {
      // The user is not logged in...
      throw Exception("User is not logged in");
    }
    final userIdFirestore = UserIdFirestore(value: userId);
```

yoannLafore:
Thank you @JoachimFavre :) 
You also greatly contributed to this PR with your great review and suggestions, this was very helpful to improve the implementation and have a clean codebase  !

lib/services/database/user_repository_service.dart
JoachimFavre: I am then curious to know if it would be interesting to add a `loggedInUserIdProvider` in `login_view_model.dart`:
```dart
final loggedInUserIdProvider = StreamProvider<UserIdFirestore?>((ref) {
  return ref.watch(firebaseAuthProvider).authStateChanges().map((user) {
    if (user == null) {
      return null;
    }

    return UserIdFirestore(value: user.uid);
  });
});
```
This code snippet was completely copied of the `userProvider` and thus raises another question in me. Are the classes `Post`, `ProfileData` and `LoginUser` still used? What is the difference between a `LoginUser` and a `UserFirestore`/`UserData`? Should there be any easy way to convert from one to another? Same question for `Post` and `PostFirestore`/`PostData`.

yoannLafore: Yes it could be a good idea to add `loggedInUserIdProvider` ! This might come useful in the future altough I currently don't have an example in mind of where we could use it. But I think that this might be beyond the scope of this PR as it concerns the view model.

I think that the difference between `LoginUser` and `UserFirestore` is the following.
- `LoginUser` represents the user from the Firebase Auth service with the data that we can get from that (the `uid` and possibly the email) (here we can only get the public data of the google account) 

- `UserFirestore` represents the user as it is stored in our database, so with all the fields that we have added (joinDate, etc.)

Regarding  the classes `Post`, `ProfileData`. I think that those are here because we needed them to allow for the people working on the UI to have a way to represent the data to display while the repositories weren't implemented.
Regarding the fact of if we should keep both or not, I think once again that this should be done in another PR as this is going to imply some modification of the UI code.

JoachimFavre (approved):
After a bit of reasoning (and a lot of help from @gruvw; thanks!), I now understand better. 

Indeed, the `Post` class will be modified when doing the overview view-model; and similarly for the other classes.

For the `loggedInUserIdProvider`, my reasoning was that @TheTexanCodeur and me will need it next sprint; and thus that it had to be done before. It however indeed makes more sense to add it in another PR; we will just need to be careful and coordinate ourselves.

This was a long journey, good job everyone (especially Yoann naturaly)! :)

PR #37, opened 2024-03-26
Static overview UI foundation

CHOOSEIT:
This pull request introduces foundational elements for the Overview UI. Fixes: [#30](https://github.com/ProximaEPFL/proxima/issues/30)

It includes two enums: `NavigationbarRoutes`, defining bottom navigation bar entries, and `TimeLineFilters`, specifying timeline filter options. 

Additionally, it introduces a sample provider named `postList` that contains a list of posts for display, along with an important class, `Post`, which outlines the structure for displayable post information.

**Note that this pull request keeps the logout button for testing purposes.**

| Overview without Posts | Overview with Posts |
|---|---|
| Overview without Posts | Overview with Posts |

PR #45, opened 2024-03-27
User profile screen 

TheTexanCodeur:
This pull request imlements a first version of the user profile screen. Main features brought by this PR are:

1. **A user profile screen displaying the personal informations of the current user**: This first implementation already consumes the user profile provider, which provides user data such as the email and the token (not displayer on the screen). Other types of data (badges, profile pictures, posts, comments,...) are not yet available. This is why the information boxes are displayed but empty. 

2. **A user profile button in the home page screen**: this button (top right) leads to the user profile screen. 

3. **Test file that checks if the screen is well displayed**: this test brings the total coverages to 88% and ensures that every element is accesssible on the phone screen.

Capture d’écran 2024-03-27 à 11 50 13


CHOOSEIT:
Hello 👋🏼 ,

I wanted to share some feedback on the UI presentation after conducting a test, even though I have not reviewed the code in detail yet. The overall design and functionality looks great !

However, I have identified a minor issue concerning the display of long email address. Currently, my address seems to overflow the container:
image

Additionally, I think that we could maybe add a little padding below the comment section:
image

Aderfish:
Hello, good job so far! I have a few comments and questions:

1. I see that your test has no assertions. I understand that the data to display is not yet available, but you could make sure that your empty widgets are indeed displayed. You could do something of the form:
  
    ```
    final badgesFinder = find.byKey(badgeRowKey);
    expect(badgesFinder, findsOneWidget);
    ```
    
    
    After having defined and set badgeRowKey in the appropriate widget. You could do this for the main elements of your page.
    
    You could also already use the data you have available, the email, and check that `testingLoginUser.email`, from the firebase user mock [file](https://github.com/ProximaEPFL/proxima/blob/main/test/component/utils/mock_data/firebase_user_mock.dart), is indeed displayed at the top of your page.

2. The final layout could use some padding on the sides on a smaller display (720x1280):

> image

gruvw:
I think this PR has conflicting concerns with the theming of the application...

It should not define and use colors and font sizes on its own, but stick to material 3 basic components and let the Theme PR handle the styling of the elements.

Another remark that is maybe more linked to the Wireframing: I think we might prefer to split posts and comments display as two tabs instead of having two different scroll views stacked on top of one another. We might want to change this to have a posts tab and a comments tab in order to have a single scroll gesture area for this page.

TheTexanCodeur:
Thanks for all your remarks/suggestions!

With this new commit I am addressing the issues pointed out by @CHOOSEIT and  @Aderfish:

-Long emails are now displayed with a fade effect  if they extand too much. 
-Borders have now padding. The screen has now a better appearance.

Capture d’écran 2024-03-28 à 10 56 48

camillelnne:
Good work!

I think the profile picture could benefit from some padding, so that it aligns with the text (user name + "User Profile).
Also, where does the pink background come from? Would be great if you could remove it.

More complicated to quantify, but I feel like the style overall clashes with @CHOOSEIT minimalistic choices for the overview page (see #37 ). I suspect this might because of the black borders and font size.

Also, please always check the appearance in dark mode, it can be surprising sometimes

I will create a commit to help you with some of the adjustments I have mentioned.

camillelnne:
Made some minor adjustments as mentioned, concerning the borders and picture radius.
I suggest we refine the mockup before modifying further and get some placeholder badges, posts and comments to see if we can get rid of the remaining black borders.
I think the app theme will need to be adjusted for colors and font size.

Definitely not an easy task!

<image src="https://github.com/ProximaEPFL/proxima/assets/101193378/efe4350c-0f04-4023-a34b-c37f4bdde87f" width=300>
<image src="https://github.com/ProximaEPFL/proxima/assets/101193378/11aabfa0-d7a8-4942-ad7a-ffccd0f7a8c7" width=300>

TheTexanCodeur:
Thanks for your commit @camillelnne !

The screen indeed looks way more like the overview page of @CHOOSEIT. 

Great work!

TheTexanCodeur:
I just commited a refractor of the user profile screen. Here are the main changes:

-One subfolder dedicated to the user info (centauri points, username , profile picture,...)

-One subfolder dedicated to the posts/comments. I decided to create a dedicated folder for those infos as they are more specific and complexe compared to other more general infos. 

This new code organisation is more clear and easier to work with. Separating the widgets will also simplify the process of a future UI rework (as decided with the group). 

Let me know what you think of it!

TheTexanCodeur:
Update about the situation of this PR:

After discussion with the team, we concluded that this particular screen needs to be reworked in the future, but we consider the task requirements for this week fulfilled. This PR was focused on the functional aspect of the screen and is thus ready to be merged. A new PR will be open to adress the change of design for this screen as it is a different issue.

camillelnne:
Great refactoring! The code is way more digest. Added the documentation for an even smoother experience :)

test/component/profile_page_test.dart
camillelnne, line 1: Like @Aderfish mentioned, I think you should for the presence of each widget: points, badges, posts and comments.
Also, why are there some tests for the login page in there?

camillelnne (changes requested):

TheTexanCodeur (commented):
I added tests that check that all widgets are displayed on the screen

camillelnne (approved):
Nice modifications!
LGTM

CHOOSEIT (changes requested):
LGTM. 

Except for the little change in the overview, this first version is good.

CHOOSEIT (approved):

Aderfish:
Good job updating the tests and only using the minimum required widgets.

PR #46, opened 2024-03-27
UI theme

camillelnne:
Fixes: #36 
This pull request sets the theme of the application:
- brightness depends on system
- color seed is indigo
- font family is `Google.poppins` (imported)

Additionally, screen rotation was locked in portrait mode.

`main.dart` and `pubspec.yaml` were updated.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1,14 +1,19 @@
 import "package:firebase_core/firebase_core.dart";
 import "package:flutter/material.dart";
+import "package:flutter/services.dart";
 import "package:hooks_riverpod/hooks_riverpod.dart";
 import "package:proxima/utils/firebase/firebase_options.dart";
 import "package:proxima/views/navigation/routes.dart";
+import 'package:google_fonts/google_fonts.dart';
 
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
   await Firebase.initializeApp(
     options: DefaultFirebaseOptions.currentPlatform,
   );
+  SystemChrome.setPreferredOrientations([
+    DeviceOrientation.portraitUp,
+  ]);
 
   runApp(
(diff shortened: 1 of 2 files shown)


yoannLafore (approved):
Hello,
Thanks for your work !

The PR looks good to me.

CHOOSEIT (approved):
Looks good to me

This is a good idea to lock the app in portrait mode.

lib/main.dart
```diff
@@ -9,6 +11,9 @@ void main() async {
   await Firebase.initializeApp(
     options: DefaultFirebaseOptions.currentPlatform,
   );
+  SystemChrome.setPreferredOrientations([
+    DeviceOrientation.portraitUp,
```
TheTexanCodeur, line 15: Great idea: this will prevent display problems and will limit the UI workload (only one orientation to take into account)

lib/main.dart
```diff
@@ -22,10 +27,21 @@ class ProximaApp extends StatelessWidget {
...
+      theme: ThemeData(
+        useMaterial3: true,
+        // Define the default brightness and colors.
+        colorScheme: ColorScheme.fromSeed(
+          seedColor: Colors.indigo,
+          brightness: brightness,
+        ),
+        fontFamily: GoogleFonts.poppins().fontFamily,
+      ),
     );
```
TheTexanCodeur, line 45: Very good point. Setting the theme/font at the top level will ensure that the whole app looks the same. We will need to be carefull in the future not to use custom colors in other widgets.

TheTexanCodeur (approved):
Overall, a great addition to the app! It's a important point/question that had to be resolved early in the project. 

The implementation is clean. I have nothing to say about it.

Good job :)

PR #47, opened 2024-03-27
Static analysis in-depth tool

### LINKED ISSUE #27: Research and add a code analysis tool

gruvw:
Fixes #27

This PR aims to find a viable alternative to SonarCloud that provides additional feedback and in depth static analysis to the Flutter/Dart project.

After conducting extensive internet research on the subject, I found a few candidates.
Unfortunately, most of the alternatives that I tried were only available as paid plans (which can get quite expensive).

I tried out the free alternatives on the https://github.com/gruvw/flutter_demo repository to check the feedback it was providing.

- codescene.io - This option turned out to not provide the best insights. It was providing an online dashboard with a few code changes metrics, but nothing that was really helpful. It could connect with the coverage command, but we already have a GitHub action that handles the coverage metric.
- codacy.com - It was marketed for Flutter and Dart projects on a few posts that I found online, but isn't really best suited to this kind of code bases in the end. It was detecting the Gradle project along with the Java files instead of actually analyzing the dart code (even after configuration).
- dcm.dev - A static analysis toolkit that helps you identify and fix problems in a Dart/Flutter code base. This tool was by far the most suited to handle code reports for this project, as it is specially crafted for analyzing Dart projects.

Therefore, this PR adds support for the [Dart Code Metrics](https://dcm.dev/) tool on this repository.

I configured a new GitHub action that checks for a few rules and metrics using `dcm` and that generates a report for every pull request suggesting some changes to the code.

Example of an automatically generated report:
image

**Note**: those changes are not mandatory and should only be seen as tooltips.

This tool does not serve the exact same purpose as SonarCloud, but it is the best alternative that I found.
We have to keep in mind that the Flutter tooling is already awesome, and it comes with a powerful static code analyzer `flutter analyze` out of the box.

**Note**: The default Flutter static code analyzer was already checked (and rules are enforced) within our CI pipeline.


yoannLafore (approved):
Thanks for your researches !

I agree that DCM should be the best alternative given what we are doing especially considering that we already have the `flutter analyze` as you mentioned.
Also, the CI integration looks good to me.

PR #38, opened 2024-03-26
Login screen

camillelnne:
This pull request implements the UI for the authentication screen. Fixes: #28.

The UI is composed of three elements:
- application logo (note: might change in the future)
- slogan
- "Sign in with Google" button

A folder `assets` was created to store the logo and future assets. Therefore the file `pubspec.yaml` was updated to reference it.

Some tests in `login_test.dart` were modified to adapt to the design:
- changed the text expected in the login button
- verify the presence of the logo
- verify the presence of the slogan

login-screen

PR #49, opened 2024-03-27
create new post screen

### LINKED ISSUE #29: Create a post screen

Aderfish:
Fixes #29,

This PR implements the new post screen.
image

The post button checks that the title and body are not empty. However it does not yet connect to the database, and there is therefore no viewmodel yet.
The settings button is not implemented yet. It will be implemented when user notifications and tag functionality will be implemented.


gruvw:
Nice work, this might require a small amount of UI code refactoring, similar to https://github.com/ProximaEPFL/proxima/pull/37#issuecomment-2024852105 and https://github.com/ProximaEPFL/proxima/pull/38#issuecomment-2022745226.
I think it could be a good idea that @CHOOSEIT or @camillelnne can assist you with that (if they want / have time), now that they have had an example on their code :)

Aderfish:
Thank you for the feedback, I did the refactoring to make the code more readable. I also changed the tests so that they only use this page and not the whole app to run. Please tell me if you find anything else that should be improved.

lib/views/pages/new_post/new_post_form.dart
```diff
@@ -0,0 +1,100 @@
...
+          Navigator.pop(context);
+        }
+      },
+    );
+
+    final settingsButton = IconButton(
+      onPressed: () {
+        // TODO open tag and notification settings overlay
+      },
+      icon: const Icon(Icons.settings),
```
TheTexanCodeur, line 75: I think it would be a better idea to put the notification settings button in the top left of the screen (in a lot if app it's his default position).

TheTexanCodeur (approved):
Good job!

The code is very clear and well structered. I added a UI design suggestion but you followed the wireframe so it's not to be addressed immediatly. I think we can go this way for now!

gruvw (approved):
LGTM, great refactoring.

PR #68, opened 2024-03-28
Fix create account UI overflow

### LINKED ISSUE #51: create user profile bug

JoachimFavre:
Fixes #51. 

More than rendering the view scrollable, I also added some minimum spacing between the widgets. 

Note that I needed to create some `ScrollableIfTooHigh` widget, to handle widgets with a flexible size (this is the case for my spacer). This was strongly inspired from [the documentation](https://api.flutter.dev/flutter/widgets/SingleChildScrollView-class.html).

The things I am least sure about and that might be interesting for you to consider:
- `ScrollableIfTooHigh` is in `utils/ui`. This makes sense in my opinion, since this is also where `CircularValue` is; but I am not completely sure.
- I tend to always use `Flexible` followed by `SizedBox` in order to have a minimum spacing between widgets, which is increased in a flexible way. I did not find something that would encapsulate the two in the same widget in the documentation, or a better way to do it; but not finding something does not mean it does not exit.

I tested it on a small screen emulator, and on my phone horizontally (commenting out the line preventing to do so in `main.dart`), which you can see hereinbelow. 

<p align="center">
<image width=30% src="https://github.com/ProximaEPFL/proxima/assets/22430066/ec4a5913-f963-487c-a4f7-9af0ce4a8a40">
<video src="https://github.com/ProximaEPFL/proxima/assets/22430066/688bebd8-5ac2-4a1b-ac0b-9ff7aee021e8">
</p>

### DIFF
--- a/lib/utils/ui/scrollable_if_too_high.dart
+++ b/lib/utils/ui/scrollable_if_too_high.dart
@@ -1,0 +1,28 @@
+import "package:flutter/widgets.dart";
+
+class ScrollableIfTooHigh extends StatelessWidget {
+  final Widget child;
+
+  const ScrollableIfTooHigh({
+    super.key,
+    required this.child,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    return LayoutBuilder(
+      builder: (BuildContext context, BoxConstraints viewportConstraints) {
+        return SingleChildScrollView(
+          child: ConstrainedBox(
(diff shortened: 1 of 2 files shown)


yoannLafore (approved):
Everything looks good to me !
Good idea to have created `ScrollableIfTooHigh` separately, this will probably be quite useful in other scenarios :+1:  
Thanks for your work !

PR #70, opened 2024-04-04
Milestone 1 application navigation (flow and UI)

CHOOSEIT:
This PR introduces several updates to the navigation flow of our application.

It includes: 

- Refactoring of the back button widget, now usable for every pages `LeadingBackButton`.
- Navigation to the new post page from the home page (feed page).
- The separation of navigation flows from login process to account creation and the home page.

This PR also include the tests and mocks associated to the navigation flows.


CHOOSEIT:
This commit fixes the CircleAvatar's InkWell that appearing rectangular on click. Now it appears circular.

test/component/login_test.dart
```diff
@@ -1,83 +1,163 @@
...
-    await tester.pumpAndSettle();
-
-    final loginButton = find.byKey(LoginButton.loginButtonKey);
-    await tester.tap(loginButton);
-    await tester.pumpAndSettle();
-
-    final createAccountPage = find.byType(CreateAccountPage);
-    expect(createAccountPage, findsOneWidget);
-
-    final logoutButton = find.byKey(CreateAccountPage.logoutButtonKey);
```
gruvw, line 75: I don't see any equivalent test for the logout functionality. I think it was completely removed in this branch, and we should not remove tested functionality if it is still a feature of the application.

Can you please make this test the logout again ? Just add a logout button either on the overview or on the profile page.

The test name says "Login and Logout using create account page".

lib/views/bottom_navigation_bar/navigation_bar_routes.dart
gruvw, line 1: We might want to move the `bottom_navigation_bar` directory to the `navigation` directory, it would make more sense.

lib/views/bottom_navigation_bar/navigation_bottom_bar.dart
```diff
@@ -6,6 +6,7 @@ import "package:proxima/views/bottom_navigation_bar/navigation_bar_routes.dart";
 /// It contains the navigation routes to the different pages.
 class NavigationBottomBar extends HookConsumerWidget {
   static const navigationBottomBarKey = Key("navigationBottomBar");
+  static const selectedIndex = 0;
```
gruvw, line 9: This should not be static const as it might change in the future. We should put it as a field of the widget and have it passed as argument to the constructor.

test/component/login_test.dart
gruvw: Actually, we should have two tests for this now, one from the overview, and one for the intermediate page. Both should logout the user in some way.

lib/views/bottom_navigation_bar/navigation_bar_routes.dart
gruvw: the same goes for `leading_back_button`

lib/views/bottom_navigation_bar/navigation_bottom_bar.dart
gruvw: Fixed by 52385d5

CHOOSEIT:
Thanks @gruvw, the new Navigation bar and structure make more sense. I added a per-page label for this new scaffold `pageLabel()` in `NavigationbarRoutes.`

gruvw (approved):

lib/views/navigation/bottom_navigation_bar/navigation_bar_routes.dart
```diff
@@ -0,0 +1,58 @@
+import "package:flutter/material.dart";
+import "package:proxima/utils/ui/not_implemented.dart";
+import "package:proxima/views/content/feed/post_feed.dart";
+import "package:proxima/views/navigation/routes.dart";
+
+/// This enum is used to create the navigation bar routes.
+/// It contains the name and icon of the routes.
+enum NavigationbarRoutes {
+  //TODO set the routes for the pages
```
JoachimFavre, line 9: I don't understand this TODO. Isn't it already done?

lib/views/pages/home/home_page.dart
```diff
@@ -22,14 +23,21 @@ class HomePage extends HookConsumerWidget {
       }
```
JoachimFavre, line 23: This should be navigateToLoginPageOnLogout(context, ref); It should moreover not be duplicated with app_top_bar.dart:23. I'd recommend to only keep the one here (calling navigateToLoginPageOnLogout naturally) and to remove the one of app_top_bar.dart

lib/views/pages/create_account_page.dart
```diff
@@ -17,20 +17,20 @@ class CreateAccountPage extends HookConsumerWidget {
...
-          },
-          icon: const Icon(Icons.arrow_back),
+    return PopScope(
+      canPop: false,
+      onPopInvoked: (bool didPop) {
+        ref.read(loginServiceProvider).signOut();
+      },
+      child: Scaffold(
+        appBar: AppBar(
+          leading: const LeadingBackButton(),
```
JoachimFavre, line 27: Why was it chosen to use a PopScope, instead of the callback from LeadingBackButton?

test/component/home/navigation_home_test.dart
```diff
@@ -0,0 +1,120 @@
...
+      (tester) async {
+    await tester.pumpWidget(emptyMockedPage);
+    await tester.pumpAndSettle();
+
+    // Check that the home page is displayed
+    final homePage = find.byType(HomePage);
+    expect(homePage, findsOneWidget);
+
+    //Click on the new post button on the home page
+    final newPostButtonText = find.byKey(PostFeed.newPostButtonTextKey);
```
JoachimFavre, line 105: Adding a `expect(newPostButtonText, findsOneWidget);` would allow to better understand the test error message in case of a potential breaking change later.

test/component/home/static_home_test.dart
```diff
@@ -84,7 +86,40 @@ void main() {
...
+  testWidgets(
+      "check correct number of navigation bottom bar elements are displayed",
+      (tester) async {
+    final homePageWidget = ProviderScope(
+      overrides: [postList.overrideWithValue(List.empty())],
+      child: const MaterialApp(
+        title: "Proxima",
+        home: HomePage(),
+      ),
+    );
```
JoachimFavre, line 106: There is a lot of code repetition in the tests, this is an example.
I don't think this is really important to fix right now, because it would require extensive refactoring, without huge interest. This time is probably best invested somewhere else for now. However, I think that keeping this in mind for the following PRs might be interesting. :)

test/component/login_test.dart
```diff
@@ -1,83 +1,202 @@
...
+    testWidgets("Login and get to home page flow", (tester) async {
+      await tester.pumpWidget(getMockedProxima());
+      await tester.pumpAndSettle();
+
+      final loginButton = find.byKey(LoginButton.loginButtonKey);
+      await tester.tap(loginButton);
+      //Needs a delay to allow the existance check to complete
+      await tester.pumpAndSettle(delayNeededForAsyncFunctionExecution);
+
+      final loginButtonStillHere = find.byType(HomePage);
```
JoachimFavre, line 65: By humble curiosity, what was the reasoning behind this variable name (same question for all the `loginButtonStillHere` variables in this test file)?

test/component/login_test.dart
```diff
@@ -1,83 +1,202 @@
...
+      );
+    }
+
+    testWidgets("Login and get to home page flow", (tester) async {
+      await tester.pumpWidget(getMockedProxima());
+      await tester.pumpAndSettle();
+
+      final loginButton = find.byKey(LoginButton.loginButtonKey);
+      await tester.tap(loginButton);
+      //Needs a delay to allow the existance check to complete
```
JoachimFavre, line 62: I don't on my computer, I can remove this delay (same for all occurences below). Is this device dependent?

test/component/login_test.dart
```diff
@@ -1,83 +1,202 @@
...
+      final logoutButton = find.byKey(AppTopBar.logoutButtonKey);
+      await tester.tap(logoutButton);
+      await tester.pumpAndSettle();
+
+      //Check that we are in the login page
+      final loginPage = find.byType(LoginPage);
+      expect(loginPage, findsOneWidget);
+    });
+
+    testWidgets("Login and get to create account page flow", (tester) async {
```
JoachimFavre, line 140: What is the difference with the test on line 89?

JoachimFavre (commented):
This is a nice PR! :)

You can see my comments below, but I also have some more general ones.

First, you moved many files to `views/content/feed`. There is however nothing in the package `views/content`, what would you consider going there? Also, why did you choose not to put those files inside `views/pages` package?

Second, I'm guessing that you renamed the class to PostFeed, for it to be more easily reusable by the profile, is it correct?

Third, I'm not very fund of the names of the tests. This is very personal, but I have to admit I sometimes feel easier to understand what a test does by reading it instead of reading its title; which is an issue. Just like the code repetition in tests, I don't think this is really worth the time to refactor, but I just think that it might be interesting for the future. :)

CHOOSEIT:
Hello, thank you for your review

> First, you moved many files to `views/content/feed`. There is however nothing in the package `views/content`, what would you consider going there? Also, why did you choose not to put those files inside `views/pages` package?

It was a change from @gruvw. It is to separate the pages that we are going to `push/pushReplacement` from the one that will be swapped within the global scaffold / navigation in the home page.

Now we use a global `Scaffold / NavigationBottomBar / AppTopBar` that is responsible to change the body of the scaffold whenever we select a route in the `NavigationBottomBar`. The idea is to store every body possible within `views/content` and use the  `NavigationbarRoutes` enum to navigate and swap the `body / content`.


> 
> Second, I'm guessing that you renamed the class to PostFeed, for it to be more easily reusable by the profile, is it correct?
> 

No it was more to separate the feed from the `Home` page. The idea is that the home is the global structure of the app, and the list of post (feed) is the content of the "Feed" page.

lib/views/navigation/bottom_navigation_bar/navigation_bar_routes.dart
CHOOSEIT: Good catch !

lib/views/pages/home/home_page.dart
CHOOSEIT: Good catch, makes also sense to keep the one in the home page as the logout button in the app bar is temporary.

test/component/home/navigation_home_test.dart
CHOOSEIT: You are right

lib/views/pages/create_account_page.dart
CHOOSEIT: That way it also handle and cancel the back arrow on android phone (to make sure that we go back to the login back only if we are log out) and keep the `LeadingBackButton` be a more general widget that is just there to pop the current page.

test/component/login_test.dart
CHOOSEIT: It dependent on the device, on my computer the test does not pass, since we are mocking the future we can be sure that it will be executed instantly, but sometimes the future is not started yet when we return from the tap on the `loginButton`.

test/component/home/static_home_test.dart
CHOOSEIT: Good catch

test/component/login_test.dart
CHOOSEIT: Good catch, I think it was for me during my test creation to debug and I forgot to rename it.

CHOOSEIT:
Thank you for you review 

> First, you moved many files to `views/content/feed`. There is however nothing in the package `views/content`, what would you consider going there? Also, why did you choose not to put those files inside `views/pages` package?

It is to separate the page that are are needed to be `push/pushReplacement` from the page that just change the body of the global scaffold that we set in `HomePage`, choosing using the `NavigationBottomBar` and the `NavigationbarRoutes` enum. 
In other word the widgets in `content` we are just swapping the body and not changing page completely.

> Second, I'm guessing that you renamed the class to PostFeed, for it to be more easily reusable by the profile, is it correct?

No it is just to separate the "Home" from the "Feed" nomenclature, the Home is the global scaffold of the app ensuring navigation between different content, and the feed is one of the possible content.


> Third, I'm not very fund of your test naming. This is very personal, but I have to admit I sometimes feel easier to understand what a test does by reading it instead of reading its title; which is an issue. Just like the code repetition in tests, I don't think this is really worth your time to refactor, but I just think that it might be interesting for you to have this in the back of your mind for the future. :)

You are right, I and will try to improve the naming thanks.

lib/views/pages/home/home_page.dart
JoachimFavre: No, I don't think we would want it in the app top bar, since it only needs to be called once in the entire navigation stack. It's nice you also removed it in your commit :)

lib/views/pages/create_account_page.dart
JoachimFavre: Oh, that's interesting! Good idea and good job :)

JoachimFavre:
Great changes! 

> It is to separate the page that are are needed to be push/pushReplacement from the page that just change the body of the global scaffold that we set in HomePage, choosing using the NavigationBottomBar and the NavigationbarRoutes enum.
> In other word the widgets in content we are just swapping the body and not changing page completely.

I am wondering if `home_content` would make more sense than `content` as a package name. What would you think?

test/component/login_test.dart
JoachimFavre: Ok, that's weird. This might become a bigger issue if tests pass on some people's computer but not on other's/the CI.
Thanks :)

JoachimFavre (approved):
LGTM now. Great job :)

PR #73, opened 2024-04-07
Mock repositories

yoannLafore:
Hello,
This pull request aims to provide the boiler plate code needed to mock the methods of the repositories.
This code is contained in the classes `MockPostRepositoryService` and `MockUserRepositoryService`.

Here is an example of how such a class can be used in a test to mock a repository method
```dart
late MockPostRepositoryService postRepository;

setUp(() {
  // Initialize the mock
  postRepository = MockPostRepositoryService();
});

test("some test", () {
  final point = ...
  final kmPostRadius = ...

  // To mock the [getNearPosts] method when it is called with particular parameters
   when(postRepository.getNearPosts(point, kmPostRadius)).thenAnswer(
        // Here, it is mocked to return an empty list
        (_) async => [], 
   );

  // From this point on, when the "postRepository.getNearPosts(point, kmPostRadius)" is called, the mock overrides
  // enters into action and an empty list will be returned
});


```

This PR also provides a class to easily generate mock data for `UserFirestore` which can be useful when testing.
This class is `MockUserFirestore` located under `test/models/database/user`.

Finally, `MockFirestorePost` has been renamed to `MockPostFirestore` as well as being moved to `test/models/database/user` for consistency.

### Important note:
For the mock of the repositories, the way the code is written can misleadingly make believe that a particular value will be returned whithout having to mock it in the tests with `when(...) ...`.
This, is not the case, if a method from the mock repository is called without having properly set up the `when(...) ...` before, an exception will be thrown.

For example, in the `MockUserRepositoryService`, we have the following override:
```dart
  @override
  Future<bool> doesUserExist(UserIdFirestore uid) {
    return super.noSuchMethod(
      Invocation.method(#doesUserExist, [uid]),
      returnValue: Future.value(false),
    );
  }
```

In our test code we could be tempted to simply call:
```dart
userRepository.doesUserExist(...);
```
expecting it to return it `false`.

This wont work and will throw an exception.
The correct way is:
```dart
when(userRepository.doesUserExist(...)).thenAnswer(
  () async {return false;}
);

userRepository.doesUserExist(...);
```

### DIFF
--- a/test/models/database/post/mock_post_data.dart
+++ b/test/models/database/post/mock_post_data.dart
@@ -8,7 +8,7 @@ import "package:proxima/models/database/user/user_id_firestore.dart";
 
 /// Helper class to create mock post data to be used in tests
 class MockPostFirestore {
-  PostFirestore createPostAt(
+  static PostFirestore createPostAt(
     PostData data,
     GeoPoint location, {
     id = "post_id",
@@ -25,7 +25,7 @@ class MockPostFirestore {
     );
   }
 
-  List<PostData> generatePostData(int count) {
+  static List<PostData> generatePostData(int count) {
     return List.generate(count, (i) {
       return PostData(
         description: "description_$i",
--- a/test/models/database/post/mock_post_data.dart
+++ b/test/models/database/post/mock_post_data.dart
@@ -6,6 +6,7 @@ import "package:proxima/models/database/post/post_id_firestore.dart";
 import "package:proxima/models/database/post/post_location_firestore.dart";
 import "package:proxima/models/database/user/user_id_firestore.dart";
 
+/// Helper class to create mock post data to be used in tests
 class MockPostFirestore {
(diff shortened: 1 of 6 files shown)


gruvw:
I think most of those newly created methods should be static. You should not need an instance of `MockPostFirestore` or similar.

gruvw:
Also, it is a good idea to mimic the directory structure of the original `lib` in the tests :+1:

yoannLafore:
Thanks for  your feedback @gruvw, you're right, the methods of `MockPostFirestore` and `MockUserFirestore` can be static.
This is corrected in commits 14fec27b7668ba436b76a9f5222392fd097e0b8f and 0887b80fc3549a7681098a6ac16a3ba05fd18c21 :+1:

Aderfish (approved):
good job, looks good to me

gruvw (approved):

PR #74, opened 2024-04-08
Overview viewmodel

yoannLafore:
Hello,
This PR aims to provide the viewmodel that will use the services in order to expose the nearby posts on the overview page.

In more details, this PR does the following:

- Rename `models.ui.Post` to `models.ui.PostOverview` because this class corresponds to a mapping of the post informations that are displayed in the overview, it does not fully represent a post (for instance, it does not contain the comments etc.). Also, this class has been made immutable because it is essentialy a dataclass.

- Implement the `HomeViewModel` that will provides the `PostOverview` that are within a radius of 100m of the current position of the user.

- Make the overview of displayed posts manually refreshable by swipping down. And if there are no post to be displayed, a refresh button is displayed.

- Provide unit and integration tests for the `HomeViewModel`.

#### Live testing
You can test the fetching of posts in the android emulator.
To do so, I have added a post at the location (37.4219983, -122.084) on the firestore database.
Thus, if you set your device location to this position, you should be able to see the post appear in the home feed as below:

The location of the post should be the default location of the emulator, but if your location isn't the same, you can change it in the parameters and import a new location in a `.gpx` file.
To import the location of the post, you can create a `post_location.gpx` file with the following content and import it:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<gpx version="1.1" creator="ChatGPT - OpenAI">
  <wpt lat="37.4219983" lon="-122.084">
    <name>Target Location</name>
    <desc>Location at latitude 37.4219983, longitude -122.084</desc>
  </wpt>
</gpx>
```

I was initially willing to directly provide this file here but the gpx type is not supported by github.


gruvw:
Hey, I just took a quick look at the code, looking good so far!

A slight note, in [home_view_model.dart](https://github.com/ProximaEPFL/proxima/pull/74/files#diff-437e6975829014d00f1e1c5f2a62c72b74e95e251ad488bfe33080ac4ba323f0), make sure to use the Viem model provider style using an AsyncNotifier like in [flutter_demo/todo_viewmodel.dart](https://github.com/gruvw/flutter_demo/blob/0fd8de6507662c23fc73ea4726f8631d57a08709/lib/viewmodels/todo_viewmodel.dart) or [profile_view_model.dart](https://github.com/ProximaEPFL/proxima/blob/50cd6e0b937b4935e39526fe0d141f3f581daaa4/lib/viewmodels/profile_view_model.dart).

yoannLafore:
Thanks for the feedback ! 
This is true, this will make the codebase more consistent. 
I have refactor the code to use an AsyncNotifier. in 020f304a08d0cb097eadcf036ca194edf560fad9.

yoannLafore:
So, I have added a refresh button for the case where there are no post to be displayed and thus the refresh by swiping down wouldn't work.

I also had to refactor the refresh logic in order to expose the loading state when fresh posts are being queried (which wasn't the case with the previous implementation).
Thus, the refresh is now implemented directly into the `HomeViewModel`.

I think that this PR is now ready for review.

test/viewmodels/home_view_model_integration_test.dart
```diff
@@ -0,0 +1,191 @@
...
+              voteScore: element.voteScore,
+            ),
+          )
+          .toList();
+
+      // The 6 first posts are under 100m away from the user and are the ones expected
+      const nbPostsInRange = 6;
+      final postPositions = List.generate(nbPosts, (i) {
+        return GeoPoint(0.0001 + i * 0.0001, 0.0001 + i * 0.0001);
+      });
```
CHOOSEIT, line 160: I think that we can make this part a bit more explicit for future readers.

```suggestion
      final postInRange = List.generate(nbPostsInRange, (i) {
        return GeoPoint(0.0001 + i * 0.0001, 0.0001 + i * 0.0001);
      });

      //Generate post positions that are not in the range.
      // The distance between [userPosition = GeoPoint(0, 0)] and a GetPoint at
      // latitude 0.0006 and longitude 0.0006 is about 0.11 km.
      final postsNotRange = List.generate(nbPosts - nbPostsInRange, (i) {
        i = i + nbPostsInRange;
        return GeoPoint(0.0001 + i * 0.0001, 0.0001 + i * 0.0001);
      });

      final postPositions = [...postInRange, ...postsNotRange];
```

test/viewmodels/home_view_model_unit_test.dart
```diff
@@ -0,0 +1,295 @@
...
+          posterUsername: owner.data.username,
+        ),
+      ];
+
+      when(postRepository.getNearPosts(point, HomeViewModel.kmPostRadius))
+          .thenAnswer(
+        (_) async => [post],
+      );
+      when(userRepository.getUser(post.data.ownerId)).thenAnswer(
+        (_) async => owner,
```
CHOOSEIT, line 266: Code duplication with lines 247 and 248
```suggestion
```

test/viewmodels/home_view_model_unit_test.dart
```diff
@@ -0,0 +1,295 @@
...
+      when(postRepository.getNearPosts(point, HomeViewModel.kmPostRadius))
+          .thenThrow(Exception("Error"));
+
+      // Refresh the posts
+      await container.read(postOverviewProvider.notifier).refresh();
+
+      // Check the actual posts
+      final asyncPosts = container.read(postOverviewProvider);
+
+      expect(asyncPosts, isA<AsyncError>());
```
CHOOSEIT, line 292: On this test, we should maybe test whether the message of the thrown exception is the expected one.

```suggestion
      const expectedErrorMessage = "Error while fetching posts";
      when(postRepository.getNearPosts(point, HomeViewModel.kmPostRadius))
          .thenThrow(Exception(expectedErrorMessage));

      // Refresh the posts
      await container.read(postOverviewProvider.notifier).refresh();

      // Check the actual posts
      final asyncPosts = container.read(postOverviewProvider);

      expect(
        asyncPosts,
        isA<AsyncError>().having(
          (error) => error.error.toString(),
          "message",
          "Exception: $expectedErrorMessage",
        ),
      );
```

CHOOSEIT (commented):
The implementation looks good to me.
I also added few optional suggestion that I think could improve the code.

It was a good idea to create `postOwnersId` as a set to reduce the number of requests on refresh. I also pretty much liked the `PostList` widget abstraction that you made to display the feed.

The implemented feature is also well tested.

Overall, good work! Thank you! 💪🏼

test/component/home/static_home_test.dart
```diff
@@ -7,12 +7,21 @@ import "package:proxima/views/pages/home/home_page.dart";
...
 void main() {
   testWidgets("static home display top and bottom bar", (tester) async {
     final homePageWidget = ProviderScope(
-      overrides: [postList.overrideWithValue(testPosts)],
+      overrides: [
+        postOverviewProvider.overrideWith(
+          () => MockHomeViewModel(
+            build: () async {
+              return testPosts;
+            },
```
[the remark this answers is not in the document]
yoannLafore: Applied in 162dcb038858c1d4e17bb39d2c2731beb6c6a08d :+1:

test/viewmodels/home_view_model_integration_test.dart
```diff
@@ -0,0 +1,191 @@
...
+      final postData = MockPostFirestore.generatePostData(1).first;
+
+      await postRepo.addPost(
+        postData,
+        userPosition,
+      );
+
+      expect(
+        container.read(postOverviewProvider.future),
+        throwsA(isA<Exception>()),
```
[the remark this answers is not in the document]
yoannLafore: You're right, this should make the code more robust.
I apply the change in 5cf56f0174c7e28cc4321f46239c549b9fd6dce1.

test/viewmodels/home_view_model_integration_test.dart
yoannLafore: Great suggestion !
This will make the test easier to understand.
Applied in 02d13f36dd00ab7f7efc8a720da36de868c38ff7.

test/viewmodels/home_view_model_unit_test.dart
yoannLafore: Correct !
I remove the duplication in 8330ba2973f4186eedd20abf0958c0a4a793e912.

test/viewmodels/home_view_model_unit_test.dart
yoannLafore: Good suggestion !
As you said, this will make sure that the right exception is thrown leading to a more robust test :+1: 

Applied in fd2e9639c0ef77cfa13fd33552841bd13b31a4ad.

yoannLafore:
Thank you for your great feedback !
Your suggestions will make the codebase cleaner and the tests more robust :+1:

CHOOSEIT (approved):

lib/views/pages/home/posts/home_feed.dart
```diff
@@ -8,48 +10,94 @@ import "package:proxima/views/sort_option_widgets/feed_sort_option/feed_sort_opt
...
+class PostList extends StatelessWidget {
+  const PostList({
+    super.key,
+    required this.emptyHelper,
+    required this.posts,
+    required this.onRefresh,
+  });
+  static const homeFeedKey = Key("homeFeed");
+
+  final Center emptyHelper;
```
gruvw, line 87: The code should never assume the type of Widget one passes as argument, as it might as well change in the future. This should be of type `Widget`.

By the way, it is not very good practice to pass helper widgets around like that and encapsulate logic that is not managed by the current Widget agent.
This code should leave the empty out of the `PostList` and handle the empty case in the parent (`HomeFeed`) widget.

gruvw: Fixed by 729756e

lib/views/pages/home/posts/home_feed.dart
```diff
@@ -8,48 +10,94 @@ import "package:proxima/views/sort_option_widgets/feed_sort_option/feed_sort_opt
...
-              "create one",
-              style: TextStyle(
-                color: Colors.blue,
+          Row(
+            mainAxisAlignment: MainAxisAlignment.center,
+            children: [
+              const Text("No post to display, "),
+              InkWell(
+                onTap: () => {
+                  //TODO: Add navigation to create post page
```
gruvw, line 32: This can be implemented now.

gruvw: Fixed by 7da89fe

gruvw:
Ok so very good work on this ViewModel!

The code demonstrates a great understanding of the overall application architecture as well as the ViewModel implementation using Flutter.

I just pushed some small changes: center the circular value, refactor using local variables and localize the concern of the empty feed widget.

Other than that, great work :rocket:

gruvw (approved):

lib/viewmodels/home_view_model.dart
```diff
@@ -1,7 +1,70 @@
...
+  /// This will put the state of the viewmodel to loading, fetch the posts
+  /// and update the state accordingly
+  Future<void> refresh() async {
+    state = const AsyncValue.loading();
+    try {
+      final posts = await build();
+      state = AsyncValue.data(posts);
+    } catch (e, stackTrace) {
+      state = AsyncValue.error(e, stackTrace);
+    }
```
JoachimFavre, line 63: https://codewithandrea.com/tips/async-value-guard-try-catch/
I'm just reading this code to get inspiration for my PR. Shouldn't a AsynValue.guard be used here?

yoannLafore: Yes, you're totally right.
I wasn't aware of the existence of this method so thank you for the suggestion !
Applied in commit 7bd15a62afb95316691e695dc6bb37feef40807a.

Issue #78, opened 2024-04-10
Posts are not updated correctly on refresh


yoannLafore:
If we are in the overview menu with some posts displayed and that then those change on the database, then we refresh. The changes are not reflected on the displayed post.

#### Steps to reproduce
- Open the app and go into the overview with a location where there are at least one post
- On the firebase console, change the title (or another field) of one of the displayed post
- Refresh the overview by swipping down
- With the current bug, the posts wont appear updated

#### Investigation
After investigation, I think that the issue arises from the way we use the `Stream` provided by the GeoFlutterFire library in the method `winthinAsSingleStreamSubscription` (that is used in our `PostRepository.getNearPosts()`).
We currently recreate the stream each time we need to fetch the posts data and then call `first()` on it to get the data.
I believe the issue comes from the fact that firestore caches some data and so when we fetch new data, even though we re-create the stream, the old data (from cache) are first putted into the stream and only after comes the new one (from server).
This would explain why we don't see the post being updated when refreshing.

#### Possible solutions
I believe the most appropriate solution is to directly expose the stream in `PostRepository` and to then adapt the `HomeViewModel` to be a `StreamNotifier`.
This should correct the bug and on top of that would have the benefit of providing live updates of the nearby posts without having to refresh (for instance, see the number of votes changing in real time). 
(Note that we would still need to refresh if we change location)

Another quicker solution that can be used to temporally deal with the bug is to disable the firestore cache.
Thus, the first element of the created stream will be the one coming from the server.
This is not an ideal solution but should be enough to provide a working product by milestone 1.

yoannLafore:
The temporary fix of disabling the firestore cache has been applied in #79.
But a proper fix is still required.

PR #79, opened 2024-04-10
fix(firestore): disable firestore cache globally

yoannLafore:
This PR aims to provide the temporary fix to #78 of disabling firestore cache.

Concretely, the firestore cache is globally disabled in the `main` function of the application.


lib/main.dart
```diff
@@ -11,6 +12,15 @@ void main() async {
   await Firebase.initializeApp(
     options: DefaultFirebaseOptions.currentPlatform,
   );
+
+  // TODO: Remove this line when the issue is correctly fixed (https://github.com/ProximaEPFL/proxima/issues/78)
```
JoachimFavre, line 16: Issue #78 will close when this PR (#79) is merged. I don't think there is something necessary to change to reflect this, but I just wanted to point it out.

JoachimFavre (commented):
LGTM! I just added a small observation that I think is important.

It's really nice removing the cache fixed the bug. Good job :)

lib/main.dart
yoannLafore: You are right, I didn't though of that.
I think what we can do is to reopen the issue #78 after this PR is merged.

TheTexanCodeur (approved):
LGTM!

We just need to keep in mind that this only a temporary fix for now. This is why i think that he todo in the code is a good idea  We won't forget about this point in the future thanks to this comment.

JoachimFavre:
> We just need to keep in mind that this only a temporary fix for now. This is why i think that he todo in the code is a good idea We won't forget about this point in the future thanks to this comment.

I think we should also put this as an issue in the backlog. This is vital for the code to be usable offline.

yoannLafore:
I agree with you, issue #78 has been put into the product backlog.

JoachimFavre (approved):
Everything looks perfect now, just make sure to keep the issue open :)

PR #81, opened 2024-04-10
New post view model

Aderfish:
This small PR allows the new post page to work and write the new post to the database.

The view model contains a simple function that receives the services and uses them to create and add the post to the database. The new post form obtains these services from its ref and passes them on when calling this function.

I have merged the navigation to make testing by hand easier, so it would make sense to merge that into main before this PR. This makes a large number of commits and files changed but those from navigation should of course be ignored here.

The tests are not ready yet, I am having trouble making a test user available in the tests. However I would be happy to already have your feedback on the view model.


yoannLafore:
The implementation logic with the repositories looks good to me, good job :+1: 
Also, good idea to check that the posts are not empty.

I just think that the view model needs some refactoring to better separate the concerns.
I took inspiration from the PR of @JoachimFavre on the account creation viewmodel #84.

Following a similar scheme, I think that we could have a class to expose the state of the post (validation errors, is posted...) that is being created. Maybe something like that:
```dart
class NewPostState {
  final String? titleError;
  final String? descriptionError;
  final bool posted;

  NewPostState({
    required this.titleError,
    required this.descriptionError,
    required this.posted,
  });

  @override
  bool operator ==(Object other) {
    if (identical(this, other)) return true;

    return other is NewPostState &&
        other.titleError == titleError &&
        other.descriptionError == descriptionError &&
        other.posted == posted;
  }

  @override
  int get hashCode {
    return Object.hash(titleError, descriptionError, posted);
  }
}
```


And then we can create a viewmodel with an AsyncNotifier and integrate your current viewmodel in it:

```dart
class NewPostViewModel extends AsyncNotifier<NewPostState> {
  @override
  Future<NewPostState> build() async {
    return NewPostState(
      titleError: null,
      descriptionError: null,
      posted: false,
    );
  }

  // Add some validation functions here
  // ...

  Future<void> addPost(String title, String description) async {
    final currentUser = ref.read(userProvider).valueOrNull;
    if (currentUser == null) {
      throw Exception("User must be logged in before creating a post");
    }

    // Validate the title and description and update the state accordingly
    // ...
    if (title.isEmpty || description.isEmpty) {
      state = AsyncData(
        NewPostState(
          titleError: title.isEmpty ? "Title must not be empty" : null,
          descriptionError:
              description.isEmpty ? "Description must not be empty" : null,
          posted: false,
        ),
      );
      return;
    }

    final currPosition =
        await ref.read(geoLocationServiceProvider).getCurrentPosition();
    final postRepository = ref.read(postRepositoryProvider);

    final post = PostData(
      ownerId: UserIdFirestore(value: currentUser.id),
      title: title,
      description: description,
      publicationTime: Timestamp.now(),
      voteScore: 0,
    );

    await postRepository.addPost(post, currPosition);

    state = AsyncData(
      NewPostState(
        titleError: null,
        descriptionError: null,
        posted: true,
      ),
    );
  }
}

final newPostStateProvider =
    AsyncNotifierProvider<NewPostViewModel, NewPostState>(
  () => NewPostViewModel(),
);
```

I think that having a ViewModel of this kind should provide a better separation of concerns as the UI won't have provide the repositories to add a post.


Regarding the UI, to manage the back navigation when the post is successfully added, we could now add the following:
```dart
    ref.listen(newPostStateProvider, (previous, state) {
      if (state.valueOrNull?.posted == true) {
        Navigator.pop(context);
      }
    });
```

And for adding a post, we only need the title and description thus we can do:
```dart
     final title = titleController.text;
     final body = bodyController.text;

     ref.watch(newPostStateProvider.notifier).addPost(title, body);
```

Also, because the state exposed by the view model is and async value, you can know when it is loading or when there is an error.
So, if necessary we can display an error message or loading while the post is being added.

But except from those refactoring, I have tried to add a post and everything went perfectly well, I was able to see the post appear on the firestore database :tada: 
This really shows that things are starting to assemble together and that we are getting closer to a functionnal product !
So, thank you for your work :+1:  !

Aderfish:
Thank you for the feedback. I fixed the problems with the tests and I followed your advice and put the validation logic into the viewmodel.

I also added a loading circle while we are trying to post to the database.

Please tell me if this looks good.

yoannLafore (approved):
Nice job !
Overall the PR looks good to me :+1: 

CHOOSEIT (approved):
LGTM !

PR #84, opened 2024-04-10
Create account view model

### LINKED ISSUE #54: Create view model for user creation

JoachimFavre:
Fixes #54.

This PR adds a view-model to the account creation page. Most importantly, it validates user inputs by imposing strong limitations on the input text and by verifying the username is not already used, and sends the new account to the database when the user clicks on the button.

I also added a `loggedInUserOverrides`, to override `firebaseAuthProvider` in order to mock a logged in user; I needed it for my tests. I moreover added `UserRepositoryService::isUsernameTaken`.

There are some issues left to consider, for the future sprints:
- The verification that the username is not already taken and the addition to the database are not atomic. This should not be a problem at our scale currently, unless facing malicious actors (see #89).
- The validation of the user inputs is only done when clicking on the button. A better interface would be to tell the user immediately when they leave the text field focus. This is however very hard to do (and one of the reasons I ended up taking way too much time on this PR), because the `username` must be validated asynchronously. An idea would be to display a loading icon as a suffix to the text field until we are able to validate it, but I could not find a way to do so without race conditions (see #88).


lib/models/ui/create_account_errors.dart
```diff
@@ -0,0 +1,40 @@
+import "package:flutter/foundation.dart";
+
+/// A dataclass handling the error status of the create account page.
+@immutable
+class CreateAccountErrors {
+  final String? uniqueUsernameError;
+  final String? pseudoError;
+  final bool valid;
```
yoannLafore, line 8: Good job ! 
I think it's a very good idea to have a separated class to represent the errors state :+1: 

I have just a small suggestion:
I believe that you are using the `valid` field to represent two different concepts:
- If the account fields are valid
- If the account has been successfully added to the database

I think that we could separate those two concepts by adding a field `created` (that is true when the account has been created on the database)
And also this way we may want to make the field `valid` a function as below:

```suggestion
  final bool created;
  bool get valid => uniqueUsernameError == null && pseudoError == null;
```

If you agree with this suggestion, it might be more explicit to rename the class `CreateAccountState`.

lib/models/ui/create_account_errors.dart
```diff
@@ -0,0 +1,40 @@
...
+  /// [uniqueUsernameError] and [pseudoError] are the error messages for
+  /// the unique username and pseudo text fields respectively. If they are null,
+  /// no error is displayed. [valid] is true if the form is valid. It is here to
+  /// differentiate from the case where the fields have no error because they have
+  /// not yet been validated from the case they have no errors because they are valid.
+  const CreateAccountErrors({
+    required this.uniqueUsernameError,
+    required this.pseudoError,
+    required this.valid,
+  });
```
yoannLafore, line 20: I wonder if we could add some default values in the constructor so that we don't have to specify all the fields if we only want to show a uniqueUsernameError for instance.

With the previous suggestion in mind, we could have:
```suggestion
  const CreateAccountErrors({
    this.uniqueUsernameError,
    this.pseudoError,
    this.created = false,
  });
```

lib/viewmodels/create_accout_view_model.dart
```diff
@@ -0,0 +1,120 @@
+import "dart:async";
+
+import "package:cloud_firestore/cloud_firestore.dart";
+import "package:hooks_riverpod/hooks_riverpod.dart";
+import "package:proxima/models/database/user/user_data.dart";
+import "package:proxima/models/ui/create_account_errors.dart";
+import "package:proxima/services/database/user_repository_service.dart";
+import "package:proxima/viewmodels/login_view_model.dart";
+
+class CreateAccountViewModel extends AsyncNotifier<CreateAccountErrors?> {
```
yoannLafore, line 10: Very good job with the view model, you perform a strong validation and the code is well documented which made it easy for me to understand :+1: 

I just believe that we don't really need to have the `CreateAccountErrors` nullable for the `AsyncNotifier`.
Conceptually, I think that having it `null` would represent the same state as:
```dart
  const CreateAccountErrors({
    uniqueUsernameError = null,
    pseudoError = null,
    created = false,
  });
```

lib/services/database/user_repository_service.dart
```diff
@@ -34,6 +34,15 @@ class UserRepositoryService {
 
     return docSnap.exists;
   }
+
+  /// This method will check if the unique username [username] is already taken
+  /// by some user
+  Future<bool> isUsernameTaken(String username) async {
+    final query = await _collectionRef
+        .where(UserData.usernameField, isEqualTo: username)
```
yoannLafore, line 42: Small suggestion, I think that we could limit the number of elements queried here.
In theory, we should have at most one but if for some reason we have a bug on the db and that a lot of accounts have the same username, limitting the query to 1 could avoid some reading cost.

```suggestion
        .where(UserData.usernameField, isEqualTo: username)
        .limit(1)
```

lib/viewmodels/create_accout_view_model.dart
```diff
@@ -0,0 +1,120 @@
...
+        valid: pseudoError == null && uniqueUsernameError == null,
+      );
+    });
+
+    // Create the account before applying the new state
+    if (newState.valueOrNull?.valid == true) {
+      final uid = ref.read(uidProvider);
+      if (uid == null) {
+        // The user is no longer logged in, so they will anyway be sent to the login page
+        state = const AsyncValue.data(null);
```
yoannLafore, line 100: If you agree with the previous suggestion on adding default parameter to `CreateAccountErrors` and making it non-nullable, here we could have:

```suggestion
        state = const AsyncValue.data(CreateAccountErrors());
```

lib/viewmodels/create_accout_view_model.dart
```diff
@@ -0,0 +1,120 @@
...
+        return;
+      }
+
+      final userData = UserData(
+        username: uniqueUsername,
+        displayName: pseudo,
+        joinTime: Timestamp.now(),
+      );
+      ref.read(userRepositoryProvider).setUser(uid, userData);
```
yoannLafore, line 109: I believe we should add an `await` to make sure that the function returns only after the user has been added.
```suggestion
      await ref.read(userRepositoryProvider).setUser(uid, userData);
```

yoannLafore (commented):
Very good job :+1: 
I have gone through your code and really liked the abstraction that you have taken to expose the various possible errors to the UI, I will definitely take inspiration for the future !

You performed a strong validation on the inputs which should keep our database sane :+1: 
Also, thank you for putting a lot of documentation.
The code is also well tested.

I have just made small suggestions below.

Overall, I really liked your approach to solve this issue and I think you did a really good job :tada:

lib/models/ui/create_account_errors.dart
JoachimFavre: This is indeed a very interesting idea. I realise that the `valid` variable does not have much purpose currently, and adding a `created` allows to reason more easily about the code.

lib/viewmodels/create_accout_view_model.dart
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    state = newState;
```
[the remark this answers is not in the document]
JoachimFavre: This would not work correctly, we would be loosing the potential errors. But I get what you mean, and I'll implement it.

lib/models/ui/create_account_errors.dart
JoachimFavre: This appeared to be useful in my factoring indeed. Fixed in 02307c481556be376b9a579679970839efef6ed6.

lib/viewmodels/create_accout_view_model.dart
JoachimFavre: Indeed, there was a mixing in my reasoning because of the original variable `valid`, which rendered the nullable state superfluous. 
I think you are correct about the fact it simplifies the code to have it non-nullable, so fixed in b0f4eb7cf45a2264956f0014196e1d89abe80789.

yoannLafore (commented):
Thank you for your work !
I think you did a great job at adapting the `CreateAccountErrors` to `CreateAccountModel` :+1:.

I have just propose the addition of a little comment below but except from that the PR looks good to me.
Great job :fireworks:

lib/viewmodels/create_accout_view_model.dart
```diff
@@ -0,0 +1,122 @@
...
+        return;
+      }
+
+      final userData = UserData(
+        username: uniqueUsername,
+        displayName: pseudo,
+        joinTime: Timestamp.now(),
+      );
+      await ref.read(userRepositoryProvider).setUser(uid, userData);
+
```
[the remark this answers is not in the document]
JoachimFavre: Good idea. Fixed in 028543dd33e1c13e5b46e00ceff99f80530f7e53.

yoannLafore (approved):

TheTexanCodeur (approved):
Great job!

After reviewing your code and the discussions with @yoannLafore, I think that this PR is ready to merge 👍🏼

PR #87, opened 2024-04-11
Fix circular value + Error message in overview

yoannLafore:
Hello,
This PR aims to fix the CircularValue widget that was wrapping it's `CircularProgressIndicator` leading to some crashes when the `CircularValue` was not integrated in a compatible parent widget.

Because of this fix, the overview is also modified to keep the `CircularProgressIndicator` centered.
This refactoring also allowed for the possibility of displaying an error message in the case where there was an error fetching the nearby posts.
Thus with this PR, if the nearby posts failed to be fetched, the error message `An error occurred` will be displayed in the overview together with a button that allows the user to refresh (retry to fetch the nearby posts).

### DIFF
--- a/lib/views/home_content/feed/post_feed.dart
+++ b/lib/views/home_content/feed/post_feed.dart
@@ -1,7 +1,6 @@
 import "package:flutter/material.dart";
 import "package:hooks_riverpod/hooks_riverpod.dart";
 import "package:proxima/models/ui/post_overview.dart";
-import "package:proxima/utils/ui/circular_value.dart";
 import "package:proxima/viewmodels/home_view_model.dart";
 import "package:proxima/views/home_content/feed/post_card/post_card.dart";
 import "package:proxima/views/navigation/routes.dart";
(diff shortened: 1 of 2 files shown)


lib/views/home_content/feed/post_feed.dart
```diff
@@ -77,6 +75,27 @@ class PostFeed extends HookConsumerWidget {
...
+              child: Center(
+                child: Column(
+                  mainAxisAlignment: MainAxisAlignment.center,
+                  children: [
+                    const Text("An error occurred"),
+                    const SizedBox(height: 10),
+                    refreshButton,
+                  ],
+                ),
+              ),
```
Aderfish, line 96: This looks okay to me, but it would be nice to mention in the title and description that you added an error message to the feed.

Aderfish (approved):
LGTM

JoachimFavre (approved):
Except for @Aderfish's comment that still needs to be resolved, LGTM :)

Good job !

yoannLafore:
Thank you for your remarks !
I have updated the title and description of the PR :+1:

PR #80, opened 2024-04-10
MS1 basic project README + build script

gruvw:
This PR is documentation related.

It adds a short description of the project and its core features along with a link to the Penpot and the architecture diagram image as required per the MS1 deliverable requirements.

**Note** - this PR also include a build script that will be used to generate the APK in a consistent way from the CI.

Fixes #71

PR #109, opened 2024-04-12
Add content permissions to release action

Aderfish:
Just a small PR to make the release action have the correct permissions work.
Currently the action fails with 403 errors.

image

gh release needs this write permission. See Permissions in https://github.com/softprops/action-gh-release.

I have not found a way of testing this github action and plan on merging changes into main until it works. I know that @gruvw tested his pr #80 on a fork of the repo, but that still did not show the problem. Act is not useful either because this action needs access to a github repo and releases. If you have other ideas to test the workflow first I would be happy to know, otherwise we will just merge changes until it works.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,6 +2,9 @@ name: Flutter Application GitHub Release
 
 on: workflow_dispatch
 
+permissions:
+  contents: write
+
 jobs:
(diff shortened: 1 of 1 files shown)


yoannLafore (approved):
The looks good to me :+1: 
I also don't know how we could perform testing on such a workflow.

PR #112, opened 2024-04-12
Fix duplicate key string in profile testing

CHOOSEIT:
This PR rename key string on the badge, post and comment card. 
These widget used the same key string, which cause a test on the profile passing when the test shouldn't.


yoannLafore (approved):
Good suggestion, thanks for your attention !
The PR looks good to me :+1:

Aderfish (approved):
Looks good to me, good job on finding this problem

PR #113, opened 2024-04-12
Make user and post mock arguments nullable

Aderfish:
When writing other tests, I found that the arguments for the methods of the mocks needed to be nullable in order to use matchers like any or captureAny. You can find more information [here](https://github.com/dart-lang/mockito/blob/master/NULL_SAFETY_README.md#manually-override-a-method-with-a-non-nullable-parameter-type).

This short PR fixes this problem for the current mocks defined in the code. Please tell me if you found something I missed.


yoannLafore (approved):
You are right, this will make writting the tests easier !
Good job finding that :+1:

PR #116, opened 2024-04-14
Add stream implementation to get the nearby posts

yoannLafore:
Hello,
This PR aims to provide the correct fix to issue #78.
The main idea behind the fix is to expose a `Stream` of posts instead of using a `Future` to query them (which was causing issues as explained in #78).

To do so, this PR does the following:

#### Firestore cache:

The firestore cached is re-enabled as this was only a temporary fix brought by #79.

#### Post Repository:

Refactor the `PostRepository.getNearPost(point, radius)` method in order to directly expose the Stream of posts near the provided location (`point`). 

In practice, a listener is set up on the database for the query and the stream will emit a new list of posts each time the query resolves into a different result. This means that if a post (within `radius` around `point`) is added/removed/modified, the `Stream` will emit a new updated `List<PostFirestore>`.

#### Home View Model:

Refactor the HomeViewModel to work with the `Stream<List<PostFirestore>>`. 

Concretely, the `HomeViewModel` has been transformed into a `AutoDisposeStreamNotifier`. This component works as follow:
- In the `build` method, we create the stream that will expose the `List<PostOverview>`.
- In the UI, the provider is watched with `final asyncPosts = ref.watch(postOverviewProvider);`. (the `AutoDisposeStreamNotifier` exposes and `AsyncValue` when watched, thus `asyncPosts` keeps the type `AsyncValue` like in the previous implementation).
- Now, each time a new `List<PostOverview>` is emitted into the stream, the `postOverviewProvider` will expose a new `AsyncData` with the new `List<PostOverview>`. This will trigger a widget rebuild and thus displaying the new posts.

This has the benefits that when a nearby post is added/removed/modifed, the overview will automatically be updated without the user needing to refresh. (For instance, this means that when the user add a post, he will see it immediately in the overview).

However, the `Stream` created in the `build` is bounded to the location of the user when the `build` was called.
This means that if the user moves, the stream will still be listening for the posts nearby the old location.
This is why still need a `refresh` method.

The `refresh` method is implemented as follows:

```dart
void refresh() {
    state = const AsyncLoading();
    ref.invalidateSelf();
  }
```

So, we start by setting the state to loading to display the circular value on the screen.
Then, we call `ref.invalidateSelf()`. This has for effect of rebuilding the provider.
Concretely, this means that the current stream is closed and that a new one is created with `build`.
During the creation of the new stream, the current user position is queried again leading to a stream listening for posts near the new user location which is what we want. 

#### Tests:

The changes described above lead to the need of refactoring the tests.
I also added added 3 UI tests of the overview to test the display of the error message on failed fetch and the calling of the `refresh` when the refresh button is pressed or the post lists is scrolled down.

### Live testing:

In order to see the impact of the modification while using the application, you have multiple options that you can test.

- You can simply go and add a post at your current location and the post should appear immediately in the overview without needing to refresh.
- You can go into the overview and add some posts. Then, in the firestore console, you can change/remove posts near the device location and you should be able to see the changes appearing in the overview almost instantly without doing anything in the app.

### Ressources:

Here are some ressources that helped me understanding the functioning of dart streams and that you might find usefull:
- Creating streams in Dart : [https://dart.dev/articles/libraries/creating-streams](url)
- Asynchronous programming: Streams : [https://dart.dev/tutorials/language/streams](url)

### DIFF
--- a/lib/viewmodels/home_view_model.dart
+++ b/lib/viewmodels/home_view_model.dart
@@ -1,3 +1,5 @@
+import "dart:async";
+
 import "package:hooks_riverpod/hooks_riverpod.dart";
 import "package:proxima/models/ui/post_overview.dart";
 import "package:proxima/services/database/post_repository_service.dart";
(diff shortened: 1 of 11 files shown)


lib/viewmodels/home_view_model.dart
```diff
@@ -6,60 +8,73 @@ import "package:proxima/services/geolocation_service.dart";
...
 
-  /// Refresh the list of posts
-  /// This will put the state of the viewmodel to loading, fetch the posts
-  /// and update the state accordingly
-  Future<void> refresh() async {
-    state = const AsyncValue.loading();
-    state = await AsyncValue.guard(() => build());
+  /// Refreshes the list of posts to display the posts that are near the new location of the user
+  /// This method should be called whenever the location of the user changes.
+  void refresh() {
```
gruvw, line 71: We might want to do something closer to the documentation here, see https://riverpod.dev/docs/essentials/side_effects#using-refinvalidateself-to-refresh-the-provider.

Maybe just calling `ref.invalidateSelf(); await future;` could work ?

gruvw:
I am somewhat concern that this change is not so user friendly...

Let's say that I am at a popular spot in a crowded area, I don't think that I want the overview with all the post to refresh every time someone close makes a new post, right ?

gruvw:
As a matter of facts, other social media sites do not automatically refresh posts content or home page (think of Instagram or Youtube), you might have stale data displayed on the home page if you already loaded it, but you only really get the update when you manually refresh the page.
Otherwise, you might confuse the user by changing what he is currently ready or looking at, and even their scroll position might change on post deletion/addition.

I thought the issue here was about making sure that when the current user creates a new post, it's home page will be able to display it but not necessarily when it is an outside person that is performing this operation.

yoannLafore:
Thanks a lot for your feedback !
After discussion with the team, we decided to no longer go with this implementation for now.
Thus for now I close this PR.

PR #120, opened 2024-04-16
Username and Centauri points display 

TheTexanCodeur:
This pull request implements a new user data viewmodel that provides more information for the different screens of the app. Here are the main features brought with this PR:

1. **Real username and displayname displayed** : Those two attributes of the user are now retrieved from firestore and are thus not mock data anymore. 

2. **First implementation of the centauri points** : User have now a "centauri points" field in the database. User have 0 points by default. For now, you can increment your points (+5) by clicking on the "settings" button in the user profile screen. The data you see and increment on the screen  is retrieved and stored in the database (not a mock example).

3. **centauri points test** : A new test checks that the "increment" feature of the centauri points is working correctly.

Capture d’écran 2024-04-16 à 10 07 18

### DIFF
--- a/lib/models/database/user/user_data.dart
+++ b/lib/models/database/user/user_data.dart
@@ -19,7 +19,7 @@ class UserData {
     required this.username,
     required this.displayName,
     required this.joinTime,
-    this.centauriPoints = 0,
+    required this.centauriPoints,
   });
 
   /// This method will create an instance of [UserData] from the
@@ -30,7 +30,7 @@ class UserData {
         username: data[usernameField],
         displayName: data[displayNameField],
         joinTime: data[joinTimeField],
-        centauriPoints: data[centauriPointsField],
+        centauriPoints: data[centauriPointsField] ?? 0,
       );
     } catch (e) {
       if (e is TypeError) {
@@ -43,13 +43,13 @@ class UserData {
     }
   }
 
-  Map<String, dynamic> addPointstoDbData(int points) {
-    return {
-      usernameField: username,
-      displayNameField: displayName,
-      joinTimeField: joinTime,
-      centauriPointsField: centauriPoints + points,
-    };
+  UserData withPointsAddition(int points) {
+    return UserData(
+      username: username,
+      displayName: displayName,
+      joinTime: joinTime,
+      centauriPoints: centauriPoints + points,
+    );
   }
 
   /// This method will create a map from the current instance of [UserData]
--- a/lib/models/database/user/user_data.dart
+++ b/lib/models/database/user/user_data.dart
@@ -12,10 +12,14 @@ class UserData {
   final Timestamp joinTime;
   static const joinTimeField = "joinTime";
 
+  final int centauriPoints;
+  static const centauriPointsField = "centauriPoints";
+
   const UserData({
     required this.username,
     required this.displayName,
     required this.joinTime,
+    this.centauriPoints = 0,
   });
 
   /// This method will create an instance of [UserData] from the
@@ -26,6 +30,7 @@ class UserData {
         username: data[usernameField],
         displayName: data[displayNameField],
         joinTime: data[joinTimeField],
+        centauriPoints: data[centauriPointsField],
       );
     } catch (e) {
       if (e is TypeError) {
@@ -38,6 +43,15 @@ class UserData {
     }
   }
 
+  Map<String, dynamic> addPointstoDbData(int points) {
+    return {
+      usernameField: username,
+      displayNameField: displayName,
+      joinTimeField: joinTime,
+      centauriPointsField: centauriPoints + points,
+    };
+  }
+
   /// This method will create a map from the current instance of [UserData]
   /// to be stored in firestore
   Map<String, dynamic> toDbData() {
@@ -45,6 +59,7 @@ class UserData {
       usernameField: username,
       displayNameField: displayName,
       joinTimeField: joinTime,
+      centauriPointsField: centauriPoints,
     };
   }
 
--- a/lib/services/database/user_repository_service.dart
+++ b/lib/services/database/user_repository_service.dart
@@ -61,9 +61,6 @@ class UserRepositoryService {
 
       transaction.update(userDocRef, updatedUserData);
     });
-    //final user = await getUser(uid);
-    // final updatedUserData = user.data.withPointsAddition(points).toDbData();
-    // await _collectionRef.doc(uid.value).update(updatedUserData);
   }
 }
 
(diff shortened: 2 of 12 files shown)


lib/models/database/user/user_data.dart
```diff
@@ -38,13 +43,23 @@ class UserData {
     }
   }
 
+  Map<String, dynamic> addPointstoDbData(int points) {
```
gruvw, line 46: This should not be a method that returns a string but a method that mutates the actual Data, something like `withPointsAddition` and return a `UserData`. Then you will call `toDbData` on this new object.

lib/viewmodels/profile_view_model.dart
```diff
@@ -1,21 +1,52 @@
...
+      );
+    }
+    final userData = await userDataBase.getUser(uid);
+
+    return ProfileData(loginUser: user, firestoreUser: userData);
+  }
+
+  @override
+  Future<ProfileData> build() async {
+    return _fetchData();
```
gruvw, line 32: If the goal here is to refresh the state by recalling the build method, the code should use another system. See `ref.invalidateSelf();` and https://riverpod.dev/docs/essentials/side_effects#using-refinvalidateself-to-refresh-the-provider.

lib/viewmodels/profile_view_model.dart
```diff
@@ -1,21 +1,52 @@
...
+  }
+
+  Future<void> addPoints(int points) async {
```
gruvw, line 35: Here I am not sure that this belongs to the view model as there will be no actual action in the profile page that will directly give you new centaury points. I think this functionality should probably stay only on the model and will be called by other actions. The centaury points will be changed based on more complex rules not yet defined.

lib/views/pages/profile/profile_page.dart
```diff
@@ -65,12 +66,18 @@ class ProfilePage extends HookConsumerWidget {
...
+                centauriPoints: value.firestoreUser.data.centauriPoints,
               ),
               actions: [
                 IconButton(
+                  key: settingsKey,
                   icon: const Icon(Icons.settings),
-                  onPressed: () {},
+                  onPressed: () {
+                    //add 5 points to the user
+                    ref.read(profileProvider.notifier).addPoints(5);
```
gruvw, line 79: This code should not be merged with the main branch, as we don't want test code to stay in the UI.
I don't think we want to have a button that just gives 5 points to any user that presses it (for any amount of times).

lib/views/pages/profile/user_info/user_account.dart
```diff
@@ -7,10 +7,14 @@ class UserAccount extends StatelessWidget {
 
   const UserAccount({
     super.key,
-    required this.userEmail,
+    required this.userName,
```
gruvw, line 10: Here, why not directly pass the `UserData` data class instead of individual attributes ?

lib/models/database/user/user_data.dart
```diff
@@ -26,6 +30,7 @@ class UserData {
         username: data[usernameField],
         displayName: data[displayNameField],
         joinTime: data[joinTimeField],
+        centauriPoints: data[centauriPointsField],
```
yoannLafore, line 33: This is only a suggestion and we might want to discuss this with the team.
But given that we are adding the centauriPoints feature on top of the existing database, this means that some accounts won't have the `centauriPoints` field in their firestore document leading to an error.
Thus, we might want to check if the value is `null` and if so consider it as 0. We would have something like that:

```suggestion
        centauriPoints: data[centauriPointsField] ?? 0,
```

lib/services/database/user_repository_service.dart
```diff
@@ -44,6 +44,13 @@ class UserRepositoryService {
         .get();
     return query.docs.isNotEmpty;
   }
+
+  /// This method will add [points] to the user with id [uid]
+  Future<void> addPoints(UserIdFirestore uid, int points) async {
+    final user = await getUser(uid);
+    final updatedData = user.data.addPointstoDbData(points);
+    await _collectionRef.doc(uid.value).update(updatedData);
```
yoannLafore, line 52: Good job in adapting the repository for the addition of centauri points :+1: 

I just have a little suggestion to use a transaction to make the update atomic.
This ensures that if the user document is modified by something else at the same time (which could happen while implementing further features) the data will stay consistent on the database.
I wonder what you and @gruvw think about this ?

The code could look like:
```suggestion
    return _firestore.runTransaction((transaction) async {
      final userDocRef = _collectionRef.doc(uid.value);

      final userDocSnap = await transaction.get(userDocRef);
      final userData = UserFirestore.fromDb(userDocSnap).data;
      final updatedData = userData.addPointstoDbData(points);
      transaction.update(userDocRef, updatedData);
    });
```

It is basically the same as yours but wrapped up in a transaction for atomicity.
Here is the documentation on firestore transaction : [https://firebase.google.com/docs/firestore/manage-data/transactions?hl=fr](url).

Note that this requires to refactor a little the `UserRepository` to keep the firestore instance in the private parameter `_firestore`. (But there is no need to change the constructor input parameters)

The constructor would simply change as follows (because `firestore` is already given as parameter to the constructor):
```dart
/* rest of constructor */, _firestore = firestore;
```

yoannLafore (commented):
Good job implementing the centauri points :+1: 
This is a nice addition that will help the user to feel more invested in the application !

I overall agree with the suggestions from @gruvw.
I have also made a few little suggestions below.
Also, I was wondering if you could add a little unit test for the `UserRepository.addPoints` in `test/services/database/user_repository_test.dart` like that we cover all the repository methods in this test file :+1: 
 
Thank you for your work ! :tada:

lib/services/database/user_repository_service.dart
gruvw: I also agreed with this suggestion, thanks for pointing it out.

TheTexanCodeur:
Hello!

I implemented your suggestions. Let me know if further changes need to be applied.

lib/services/database/user_repository_service.dart
```diff
@@ -47,9 +49,21 @@ class UserRepositoryService {
...
+      final userData =
+          UserData.fromDbData(userDocSnap.data() as Map<String, dynamic>);
+
+      final updatedUserData = userData.withPointsAddition(points).toDbData();
+
+      transaction.update(userDocRef, updatedUserData);
+    });
+    //final user = await getUser(uid);
+    // final updatedUserData = user.data.withPointsAddition(points).toDbData();
+    // await _collectionRef.doc(uid.value).update(updatedUserData);
```
yoannLafore, line 66: I just believe you forgot to remove those comments
```suggestion
```

yoannLafore (commented):
Thank you for bringing the changes !
The PR looks good to me.

I have just made two minor formatting suggestion below.

Thank you for your good work !
Great job :tada:

yoannLafore (approved):
Thank you for the modifications :+1: 
Everything looks good to me :tada:

PR #121, opened 2024-04-16
User posts query

### LINKED ISSUE #104: Create model function to get the posts of a user

gruvw:
A new query functionality on the model aspect that allows any viewmodel to retrieve the lists of posts of a given user.

This functionality will prove itself particularly useful for the viewmodel of the profile page, where we want to display all the posts of a given user.

Fixes #104

### DIFF
--- a/test/models/database/post/mock_post_data.dart
+++ b/test/models/database/post/mock_post_data.dart
@@ -45,7 +45,7 @@ class MockPostFirestore {
         ownerId: userId,
         title: "title",
         description: "desciption",
-        publicationTime: Timestamp.now(),
+        publicationTime: Timestamp.fromMicrosecondsSinceEpoch(1000000),
         voteScore: Random().nextInt(100),
       ),
     );
--- a/test/models/database/post/mock_post_data.dart
+++ b/test/models/database/post/mock_post_data.dart
@@ -1,3 +1,5 @@
+import "dart:math";
+
 import "package:cloud_firestore/cloud_firestore.dart";
 import "package:geoflutterfire2/geoflutterfire2.dart";
 import "package:proxima/models/database/post/post_data.dart";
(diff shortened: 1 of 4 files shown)


gruvw:
BTW, I also think that there is a problem here as the tests are passing locally but not on the CI... idk why yet, I will investigate this issue.
Meanwhile, can somebody confirm that all the tests on this branch pass on their machine ? ty

yoannLafore:
On my machine, the tests run correctly with `flutter test`.
However, I get the same error as the CI when I run the workflow with `act`.

test/models/database/post/mock_post_data.dart
```diff
@@ -25,6 +27,30 @@ class MockPostFirestore {
...
+      ),
+      location: PostLocationFirestore(
+        geoPoint: location,
+        geohash: point.hash,
+      ),
+      data: PostData(
+        ownerId: userId,
+        title: "title",
+        description: "desciption",
+        publicationTime: Timestamp.now(),
```
yoannLafore, line 48: I just think that we may want to create the timestamp from a predefined value.
Like that its value doesn't change between runs and the tests operate on the same data.
```suggestion
        publicationTime: Timestamp.fromMicrosecondsSinceEpoch(1000000),
```

test/services/database/post_repository_test.dart
```diff
@@ -213,5 +214,36 @@ void main() {
...
+
+      expect(actualPosts1.length, 2);
+      expect(actualPosts1.length, actualPosts2.length);
+
+      for (final post in actualPosts1) {
+        expect(post.data.ownerId.value, userId1.value);
+      }
+      for (final post in actualPosts2) {
+        expect(post.data.ownerId.value, userId2.value);
+      }
```
yoannLafore, line 246: Nice test :+1: 
I believe we can make it a little more concise like that:

```suggestion
      expect(actualPosts1, unorderedEquals([postsData1, postsData2]));
      expect(actualPosts2, unorderedEquals([postsData3, postsData4]));
```

yoannLafore (approved):
The PR looks good to me :+1: 
I have just made two little suggestions below.

Overall, great job :tada: 
Thanks for your good work !

test/models/database/post/mock_post_data.dart
gruvw: Applied in f6c7890

test/services/database/post_repository_test.dart
gruvw: Applied in b06a70f

CHOOSEIT (approved):
Clean, great job !

LGTM !

PR #122, opened 2024-04-16
Clean tests

### LINKED ISSUE #95: Clean-up test files
During the implementation of the core features of the applications (models, viewmodels, views), a lot of tests have been written in a pretty independent way.
This lead to some tests ending up implementing the same (or very similar) mock classes that should be regrouped to make the codebase more readable.

In this issue, we should:
- Make global variables for the mock/overrides that are used in multiple tests
- Ensure that the various tests are written in a similar manner and follow a consistent approach to testing

camillelnne:
Fixes #95. 

This pull request aims to enhance the structure and clarity of our test suite, making it easier to maintain and setting a standard for future tests.

Previously, our test files suffered from inconsistencies in setup processes, and had some repeated code, which made them hard to maintain and extend.

Features:
- **Naming convention:** In each UI test file, proper rendering of widgets is now checked in the group `"Widgets display"`. Each test is named `"Display <component>"`.
- **Mocked Pages:** There is only one mockedPage per page displayed in each test, reducing redundancy.
- **Initialization in `setUp()`:** Global variables such as the mocked pages and firestore are all declared as `late` and initialized for each test in `setUp()`.
- **Renaming Tests:** Test names have been updated to reflect their purpose more clearly.
- **Refactored Code:** Extracted common code into new methods or removed duplications altogether.

All tests have been executed to confirm that they behave as expected.


camillelnne:
Waiting for all the PRs to be merged on main before merging mine, to centralize conflicts.

test/component/home/navigation_home_test.dart
```diff
@@ -10,134 +10,140 @@ import "package:proxima/views/navigation/routes.dart";
...
+
+      //Go back to the home page
+      await tester.tap(find.byKey(LeadingBackButton.leadingBackButtonKey));
+      await tester.pumpAndSettle();
+
+      //Check that the home page is displayed
+      expect(homePage, findsOneWidget);
+    });
+
+    testWidgets("From empty feed, flow to create a post using bottom bar",
```
TheTexanCodeur, line 81:
```suggestion
    testWidgets("From empty feed, flow to create a post using bottomBar",
```

TheTexanCodeur (commented):
Great work!

Adding consistency in our test structure was really necessary. This common organisation makes it easier for dev that didn't write the tests to understand them and write additional ones. 

I just made a small comment but overall LGTM!

TheTexanCodeur:
I renamed the ProviderScope with a user logged. The two ProviderScope are now easier to differentiate when reading the tests.

test/component/login_test.dart
```diff
@@ -162,48 +178,5 @@ void main() {
       final loginPage = find.byType(LoginPage);
       expect(loginPage, findsOneWidget);
     });
-
-    testWidgets("Login to Create Account Page to Home Page flow",
```
JoachimFavre, line 166: By humble curiosity, why was this test deleted?

test/component/new_post_test.dart
```diff
@@ -19,45 +18,51 @@ import "../services/mock_geo_location_service.dart";
...
+    setupFirebaseAuthMocks();
+
+    postRepository = MockPostRepositoryService();
+    geoLocationService = MockGeoLocationService();
+
+    mockedNewPostPage = ProviderScope(
+      overrides: [
+        ...firebaseMocksOverrides,
+        postRepositoryProvider.overrideWithValue(postRepository),
+        geoLocationServiceProvider.overrideWithValue(geoLocationService),
```
JoachimFavre, line 37: I found that `MockGeoLocaitonService` is defined in `geolocation_service_test.dart`. I think it would make sense to define in its own file, together with its override, and to use this here instead.

JoachimFavre (commented):
You have made some very important changes, this will simplify our life a lot for future testing, thank you so much!

I have some comments below. Don't be mistaken by their number, I found what you did overall very good. Most of them should be very quick to fix.

I also have another comment. `test/component/utils/home/mock_data.dart` was renamed to `test/components/utils/mock_data.dart`. I am curious, why not choosing `test/components/mock_data.dartdirectly`? What would you consider going inside `utils/`, it is currently empty?
In fact, I think it would even make sense to move `mock_data.dart` to the `test/viewmodels/home` package, together with the `mock_home_view_model`. 

Moreover, I have one bigger comment to make, about overrides. I think it is worth considering to refactor those too in this PR. I made some comments about them, but this is more general. Some mocks define overrides, others don't; and I think it could be nice to uniformise this. This means multiple things:
- If there exists an override, it should be used. This is the case of many `postOverviewProvider.overrideWith(() => MockHomeViewModel())`, which should be replaced by `mockEmptyHomeViewModelOverride`.
- If there does not exist an override, we should really consider creating one if some other test might want to use it. I am for instance thinking of `        postOverviewProvider.overrideWith(() => MockHomeViewModel(build: () async => testPosts, ), )`.
- We should consider to find a better way to list those overrides. They are currently very spread out, and it is very hard to find all of them. We could consider putting all of them in some package `overrides`, but this might loose a lot of information on models, view-models and views. Then, it might be interesting to have one `override` package for each, i.e. `model/override`, `view_model/override` and `view/override`. This is however might however be a bit strange too. Right now, the idea I find best would be to change the file organisation in tests, to have a `views/`, `viewmodels/`, and `models/` and `mocks/` packages, where the three first have tests, and the fourth one has mocks. I am sorry I don't have a more definite answer, I'll try to keep thinking about this; this is definitely the hardest thing I am commenting here.

test/component/home/navigation_home_test.dart
```diff
@@ -10,134 +10,140 @@ import "package:proxima/views/navigation/routes.dart";
...
+          find.byKey(LeadingBackButton.leadingBackButtonKey);
+      expect(leadingBackButton, findsOneWidget);
+      await tester.tap(leadingBackButton);
+      await tester.pumpAndSettle();
+
+      //Check that the home page is displayed
+      expect(homePage, findsOneWidget);
+    });
+
+    testWidgets(
```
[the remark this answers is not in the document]
camillelnne: Good catch, I can add it.

camillelnne:
> You have made some very important changes, this will simplify our life a lot for future testing, thank you so much!
> 
> I have some comments below. Don't be mistaken by their number, I found what you did overall very good. Most of them should be very quick to fix.
> 
> I also have another comment. `test/component/utils/home/mock_data.dart` was renamed to `test/components/utils/mock_data.dart`. I am curious, why not choosing `test/components/mock_data.dartdirectly`? What would you consider going inside `utils/`, it is currently empty? In fact, I think it would even make sense to move `mock_data.dart` to the `test/viewmodels/home` package, together with the `mock_home_view_model`.
> 
> Moreover, I have one bigger comment to make, about overrides. I think it is worth considering to refactor those too in this PR. I made some comments about them, but this is more general. Some mocks define overrides, others don't; and I think it could be nice to uniformise this. This means multiple things:
> 
> * If there exists an override, it should be used. This is the case of many `postOverviewProvider.overrideWith(() => MockHomeViewModel())`, which should be replaced by `mockEmptyHomeViewModelOverride`.
> * If there does not exist an override, we should really consider creating one if some other test might want to use it. I am for instance thinking of `postOverviewProvider.overrideWith(() => MockHomeViewModel(build: () async => testPosts, ), )`.
> 
> For a future task (you have already done a lot for this one!), we should consider to find a better way to list those overrides. They are currently very spread out, and it is very hard to find all of them. This is actually a very deep task, since I feel it requires refactoring the whole file tree in the `test` package. My vision would be to add a `mocks` package (for instance, renaming the `service` one), and to have `views`, `viewmodels` and `models` test that contain their respective tests, and only those tests. The `mocks` package would contain all the mock classes, mock data and mock data generator.

Hi @JoachimFavre, thanks for the review!
You made some very good points :)

Indeed there are some still come inconsistencies for the overrides. It was something I had noticed and tried to fix, before realizing that it was too much work for one sprint (some overrides differ in parameters, causing bugs) and decided to focus on the general structure of each component test files.
So not to make the matter worse, I didn't delete any overrides, nor did I used them in other files, so that it can be decided later, later whether to centralize or not.
As discussed with our current @TheTexanCodeur, the decision will be discussed in Sprint meeting and the consequent task will be added to the backlog.

Will implement your other suggestions!

test/component/home/navigation_home_test.dart
```diff
@@ -10,134 +10,140 @@ import "package:proxima/views/navigation/routes.dart";
...
-        matching: find
-            .byType(NavigationDestination)
-            .at(NavigationbarRoutes.addPost.index),
-      ),
+    emptyMockedPage = ProviderScope(
+      overrides: [
+        postOverviewProvider.overrideWith(
+          () => MockHomeViewModel(),
+        ),
+      ],
```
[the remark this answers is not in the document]
camillelnne: This will go in another sprint task dedicated to overrides and mocked pages :)

JoachimFavre:
Yes, this decision makes a lot of sense! Your PR is already +500/-500, so I agree that, in fact, adding more would be too much. Thank you for pointing it out :D

test/component/home/navigation_home_test.dart
camillelnne: Will chose bottom bar, so that we refer to the type of component in general, instead of bottomBar that is more suited for a variable name.
Sorry @TheTexanCodeur x)

test/component/home/navigation_home_test.dart
```diff
@@ -10,134 +10,140 @@ import "package:proxima/views/navigation/routes.dart";
...
-    //Check that the home page is displayed
-    expect(homePage, findsOneWidget);
+    nonEmptyMockedPage = ProviderScope(
+      overrides: [
+        postOverviewProvider.overrideWith(
+          () => MockHomeViewModel(
+            build: () async => testPosts,
+          ),
+        ),
+      ],
```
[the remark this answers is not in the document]
camillelnne: This will go in another sprint task dedicated to overrides and mocked pages :)

test/component/new_post_test.dart
camillelnne: This will go in another sprint task dedicated to overrides and mocked pages :)

test/component/login_test.dart
camillelnne: It's because it is already contained in the test "Login flow to HomePage and log out" (and in the "Widgets display" group for the slogan and button testing).

JoachimFavre: Indeed, it was redundant with other tests. Well seen!

JoachimFavre (approved):
LGTM, thank you for all your work on this PR!

PR #124, opened 2024-04-16
Fix flutter test CI

### LINKED ISSUE #123: Fix flutter test CI

gruvw:
Investigating and trying to fix the CI issue on testing that does work locally but fails on CI.

**Note**: this PR also bumps the version of the project to correspond to the current milestone/sprint.

Fixes #123

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -2,10 +2,11 @@ name: proxima
 description: "A new Flutter project."
 
 publish_to: "none"
-version: 0.1.3
+version: 0.2.1
 
 environment:
   sdk: '>=3.3.1 <4.0.0'
+  flutter: 3.19.3
 
 dependencies:
   flutter:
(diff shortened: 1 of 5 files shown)


gruvw:
Ok so after quite a lot of time spent locally trying to reproduce this issue and find the root cause, I think I finally solved it!

It was both link to the flutter version used in the CI that was not fixed and another breaking change push by firebase_core without us expecting it.

Once this gets merged into main just pull main into your feature branch and the CI issue should be solved.
Reopen #123 otherwise and ping me.

camillelnne (approved):
Thank you very much for the fix!
LGTM

pubspec.yaml
```diff
@@ -14,7 +15,7 @@ dependencies:
   hooks_riverpod: ^2.4.10
   flutter_hooks: ^0.20.5
   nanoid: ^1.0.0
-  firebase_core: ^2.27.1
+  firebase_core: 2.27.1
```
camillelnne, line 18: Just curious, why is locking the version of firebase_core necessary ?

gruvw: Because the new update contains a breaking change that might not be stabilized and could change in the future.
Apparently, they ran into some issues on their library and committed a code that was push to everyone as a minor patch, but wasn't one in reality.

In the meantime, we lock it and see what will happen to the library in the future.

yoannLafore (approved):
Thanks for bringing the fix !
The PR looks good to me :+1:

PR #125, opened 2024-04-16
Post upvotes model

### LINKED ISSUE #108: Create model for upvotes/downvotes

JoachimFavre:
Fixes #108.

This PR implements the `PostUpvoteRepositoryService`. It has two methods, one for getting the current upvote state of some user on some post, and one for setting it. They are both atomic; they use Firestore transactions.

Since this is the first time I work with Firestore transactions, I have made sure to test this in details. Actually, most of this PR (~300 lines) is tests. EDIT: There are much fewer tests now, see [the rest of the discussion](https://github.com/ProximaEPFL/proxima/pull/125#issuecomment-2061300867).

Note that, a bit as a side effect, I also had to modify the `PostRepositoryService` so that its `addPost` method also returns the ID of the newly created post. This simplifies my life in testing, and I think it can be useful in other cases.


test/services/database/post_upvote_repository_test.dart
```diff
@@ -0,0 +1,302 @@
...
+    () {
+      late PostFirestore post;
+      const userId = UserIdFirestore(value: "user_id");
+
+      setUp(() async {
+        final id = await postRepository.addPost(postData, postLocation);
+        post = await postRepository.getPost(id);
+      });
+
+      void testActions(List<UpvoteState> states) async {
```
gruvw, line 81: Is there any reason why this function is marked as async ?

lib/services/database/post_upvote_service.dart
```diff
@@ -0,0 +1,131 @@
...
+  /// list of users who downvoted the post with id [postId]
+  CollectionReference<Map<String, dynamic>> _downvotersCollection(
+    PostIdFirestore postId,
+  ) {
+    return _postDocument(postId).collection(downvotersSubCollectionName);
+  }
+
+  /// Returns the upvote state of the user with id [userId] on the post with id [postId]
+  /// This is done atomically.
+  Future<UpvoteState> getUpvoteState(
```
gruvw, line 49: This will require modifying our current viemmodel for displaying a post, as for the moment we only have access to the displaying of the upvote count but nothing was made to accommodate the state of the upvote for a given post, for the current logged-in user.

I am not yet sure if we should create a new separate `PostUpVoteViewModel` or have it as a new attribute inside the `PostOverview`.
It is not really the goal of this PR anyway, but it is definitely something to keep in mind for future integration.

lib/services/database/post_upvote_service.dart
```diff
@@ -0,0 +1,131 @@
...
+import "package:proxima/models/database/post/post_firestore.dart";
+import "package:proxima/models/database/post/post_id_firestore.dart";
+import "package:proxima/models/database/upvote_state.dart";
+import "package:proxima/models/database/user/user_id_firestore.dart";
+import "package:proxima/services/database/firestore_service.dart";
+
+/// This repository service is responsible for handling the upvotes of posts
+class PostUpvoteRepositoryService {
+  /// The name of the subcollection that contains the list of users who upvoted
+  static const upvotersSubCollectionName = "upvoters";
```
gruvw, line 13: Is there a reason behind the choice of making two separate subcollections `upvoters` and `downvoters` instead of doing only one called `voters` and have a boolean flag field in it to know if it was an upvote or a downvote ? I feel like the late would be faster (only one request) and probably would facilitate the implementation.

lib/services/database/post_upvote_service.dart
```diff
@@ -0,0 +1,131 @@
...
+
+  /// The name of the subcollection that contains the list of users who downvoted
+  static const downvotersSubCollectionName = "downvoters";
+
+  final FirebaseFirestore firestore;
```
gruvw, line 18: I think this field should be private. Take a look at other Database related services like `UserRepositoryService` for the constructor.

gruvw:
The code looks good to me, and it fulfills the requirements for this task.

I think the modification to [post_repository_service.dart](https://github.com/ProximaEPFL/proxima/pull/125/files#diff-4897a6964c038161589579c50322bda5ebf5a0a5e88564b0e08ec4d792f8443a) makes a lot of sense and should not cause any trouble.

I just have a problem with the testing that goes with this PR: I think the tests are way more extensive and complex than they should be. I understand that this has to do with the Firestore transactions (as you stated), but I really feel the complexity of those tests makes them hard to maintain without any real added benefits. Furthermore, I don't think we need the random testing mechanisms for simple unit tests in this project as we should not expect weird behavior (as we could in complex compression algorithm for example). I think the code for the tests as it is, will be hard for others to understand in depth and therefore make it hard for anyone else than you to work on this part of the code. It introduces some coupling between who wrote this code and who needs to work on it in the future if we decide to change it at some point.

On top of that, those unit tests are running on a mock of Firestore and not on the actual implementation, meaning that testing the actual order and transaction atomicity does not really make sense (as they are run on a local, in order, same process DB mock).

Overall, I think the tests that we write should stay simple ([KISS](https://en.wikipedia.org/wiki/KISS_principle)) in an effort to facilitate fast iteration and changes when they arrive.

Tell me what do you think about that ?

yoannLafore:
Hello,
Thanks for your great work !
I just  wanted to pin here the discussions that we got in person regarding the code.

In the function `setUpvoteState`, the method `getUpvoteState` is called within the transaction to get the state of the vote before setting the vote.
The issue was that `getUpvoteState` was already in itself running a transaction to get the state.
But the transactions from `setUpvoteState` and `getUpvoteState` had no link between them and thus the method `setUpvoteState` could in fact not be atomic.

Thus, a possible fix for this issue is to add a `Transaction transaction` parameter to the `getUpvoteState` method. This way, the `getUpvoteState` can perform its operations on the same transaction making `setUpvoteState` effectively atomic.

The prototype for `getUpvoteState` could look like:
```dart
Future<UpvoteState> getUpvoteState(
    UserIdFirestore userId,
    PostIdFirestore postId,
    Transaction? transaction
  )
```

test/services/database/post_upvote_repository_test.dart
yoannLafore, line 1: Good initiative to test the atomicity of the operation :+1: 
However, I am not sure that the implementation of `FakeFirestore` is done to support parallelism and concurrent operations under the hood.
It is likely that the read/writes are done synchrounisly in there implementation.
But I have currently no real way to find out if the concurrency is present `FakeFirestore` or not.

However, I have tried to remove all transactions from the code and to run the tests again (3 times) and they all passed each time.
Thus, I am not sure if we are able to detect non-atomicity.

I currently cannot find a reasonnable way to test the atomicity of those operations but I think that for now it might not be necessary as those operations are quite small.

yoannLafore (commented):
So, I have gone through your code again and the implementation looks good to me :+1: 

The code is also (very) well tested and there is a good documentation which makes it easy to read through.
I really like that you tried to test the atomicity of operations. I'm just not sure that we can really do it with `FakeFirestore`.

I have just made a few suggestions below.

Great job and thank you for your work :tada:

test/services/database/post_upvote_repository_test.dart
JoachimFavre: No indeed, fixed in eadff8fc2ad6c2e54b8b9c12de364be006f46d9c.

lib/services/database/post_upvote_service.dart
JoachimFavre: I'm guessing this will go in the viewmodel of the overview, but we will indeed see later! :)

lib/services/database/post_upvote_service.dart
JoachimFavre: My original idea was that I guessed it would simplify my implementation, and that it might make some operations faster because the list we would access would be smaller. After some more reasoning, I think you are right and this is fixed in 003cd5dd0a8097512a10ae604565a947d5c00440.

lib/services/database/post_upvote_service.dart
JoachimFavre: Definitely indeed, well seen! :)
Fixed in c65785d466d56d027069644486317e04ced3f76f.

JoachimFavre:
> Hello, Thanks for your great work ! I just wanted to pin here the discussions that we got in person regarding the code.
> 
> In the function `setUpvoteState`, the method `getUpvoteState` is called within the transaction to get the state of the vote before setting the vote. The issue was that `getUpvoteState` was already in itself running a transaction to get the state. But the transactions from `setUpvoteState` and `getUpvoteState` had no link between them and thus the method `setUpvoteState` could in fact not be atomic.
> 
> Thus, a possible fix for this issue is to add a `Transaction transaction` parameter to the `getUpvoteState` method. This way, the `getUpvoteState` can perform its operations on the same transaction making `setUpvoteState` effectively atomic.
> 
> The prototype for `getUpvoteState` could look like:
> 
> ```dart
> Future<UpvoteState> getUpvoteState(
>     UserIdFirestore userId,
>     PostIdFirestore postId,
>     Transaction? transaction
>   )
> ```
This was indeed a great observation! This is fixed in 320d74e112c90016cf7bf59c311c5b072e7e8ec6.

test/services/database/post_upvote_repository_test.dart
```diff
@@ -0,0 +1,292 @@
...
+          posts.add(await postRepository.getPost(id));
+        }
+      });
+
+      for (final newState in UpvoteState.values) {
+        test(
+          "User can apply action ${newState.name} on $nPosts different posts",
+          () async {
+            for (final post in posts) {
+              assertPostUpvoteState(userId, post.id, UpvoteState.none);
```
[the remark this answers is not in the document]
JoachimFavre: Indeed! I'm guessing this still worked since it does not have any side effect, but it may have weird non-deterministic implications. Well seen! :)
Fixed in 89c870729ae4119d4d6b7a5e6230c7244bc7f8d0.

JoachimFavre:
You both raised some very important concerns concerning my tests on data races. I have done some more intensive testing, and I came up with the following observations:
- It is possible to create and detect data races using `MockFirebaseFirestore`.
- `MockFirebaseFirestore` does not handle transactions atomically, they just run its content as if it were any other sequence of instructions (see [their code](https://github.com/atn832/fake_cloud_firestore/blob/6164f7f0e9b28ee41853cd6859d8faba4a7f130f/lib/src/fake_cloud_firestore_instance.dart#L116)). This could be fixed by locking a mutex when running a transaction, for instance using [the `synchronized` package](https://pub.dev/packages/synchronized). This would however make all instructions in a transaction atomic, which is correct. For instance, putting a database access in a transaction without declaring it as part of the transaction is [a mistake I made](https://github.com/ProximaEPFL/proxima/pull/125#issuecomment-2060922176), which would not be detected by using such a mutex.

This means that we are indeed able to detect dataraces for bad code, but we are not able to differentiate bad code from good code; rendering any tests about data races dangerous (they might fail when the code is correct, or pass when the code is incorrect and give incorrect confidence). Thus, I removed my tests about dataraces in 26beafb7fb23275a169b729ce35d86702694094a. 

This is not very important since those tests were removed, but to answer to @yoannLafore's concern _"I have tried to remove all transactions from the code and to run the tests again (3 times) and they all passed each time"_, this appears to be because my code does not have a datarace on these particular tests (my bad). To make a datarace show up, we would need the same user to make many actions on the same post at the same time. One could therefore ague that I could keep the test I removed, but I disagree since, if we change the way this is implemented, there might be a datarace appearing in it.

In the very long run, we might be interested in using [Firebase's emulator suite](https://firebase.google.com/docs/emulator-suite), which is the way they propose for extensive testing. If you think it is interesting, I will add it to the backlog. I will let the product owner decide, but I am guessing this will probably not be very important since, as Lucas pointed out in an informal discussion, small data races would not be very sensitive in our context; we are not building a medical app to paraphrase his words.

Finally, to make everything simpler, I added more documentation to the tests in 5b4785b9ff1674e4e65c2fb5389832e79e6f1f6f, to make them more understandable. Tell me if you think that they are still to complex and that I should simplify some of them, or remove some other tests.

yoannLafore (commented):
Thanks a lot for bringing these changes and for the time you put experimenting with those tests !
This will really help to get a robust codebase.

I have gone through the PR and pretty much everything looks good to me. 
I just made a quick suggestion below.
Except from that I believe the PR is ready for merge :+1: 

Thank you for your great work and the effort you put with those tests !
Overall, great job :tada:

gruvw (approved):
Looks good to me now. Thank you for taking the suggestions into account :)

JoachimFavre:
Thank you all!

Just a quick remark for the future selves reading the commit tree, I used GitHub's feature to apply @yoannLafore's proposed commit, but I misclicked it and I was not able to modify the commit message. This is why 4ccc6c83ee095224820561721bc871f8a3d1e706 has an aweful commit message, it should be instead "fix(UpvoteService): Re-throw errors on VoteFirestore creation"

yoannLafore (approved):

Issue #107, opened 2024-04-12
Handle error case on circular value


TheTexanCodeur:

Aderfish:
Show an error popup if an error occured

PR #127, opened 2024-04-17
Circular value error handling

Aderfish:
Addresses #107 
This PR adds error handling to CircularValue. This allows us to reuse this component anywhere we have an AsyncValue, and automatically handle errors.

When the AsyncValue becomes and error, a fallback widget is built in place of the CircularValue, and an error popup is shown with a string representation of the error. The default fallback widget is an empty widget. The default fallback is empty because the circular value might be used in small spaces, where there is no room to show any text.

PostFeed has been adapted to use CircularValue, and to show the previous error message with a refresh button as a fallback widget.
NewPostViewModel has been updated with a guard, to catch errors and store them in the AsyncValue state.

You can see the error handling in action by disabling the position service and refreshing the home page or trying to add a new post.

After refreshing the new post page, first we get a popup:
image

And then as fallback the same screen as before:
image

On the new post page, after clicking on post with location off we get the popup:
image

And then the empty default fallback:
image

In later PR's we can imagine creating custom popups for different exception types. We will also have to think about handling the case when we do not have internet connection. Currently the post feed returns an empty feed without throwing an error, and the new post page hangs forever without throwing an error.

### DIFF
--- a/test/component/utils/ui/circular_value_test.dart
+++ b/test/component/utils/ui/circular_value_test.dart
@@ -1,0 +1,72 @@
+import "dart:math";
+
+import "package:flutter/material.dart";
+import "package:flutter_test/flutter_test.dart";
+import "package:hooks_riverpod/hooks_riverpod.dart";
+import "package:proxima/utils/ui/circular_value.dart";
+import "package:proxima/utils/ui/error_alert.dart";
+
+void main() {
+  Widget testCircularValue(AsyncValue<void> value) => CircularValue(
+        value: value,
+        builder: (context, data) => const Text("Completed"),
+        fallbackBuilder: (context, error) => const Text("Strange Error"),
+      );
+
+  testWidgets(
+      "CircularValue should show CircularProgressIndicator when loading", (
+    tester,
+  ) async {
+    await tester.pumpWidget(
+      ProviderScope(
+        child: MaterialApp(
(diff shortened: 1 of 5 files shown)


test/component/utils/ui/circular_value_test.dart
```diff
@@ -0,0 +1,71 @@
...
+    tester,
+  ) async {
+    await tester.pumpWidget(
+      ProviderScope(
+        child: MaterialApp(
+          home: testCircularValue(const AsyncValue.data(null)),
+        ),
+      ),
+    );
+
```
TheTexanCodeur, line 39: We might want to wait for the Widget to settle before we check it.
```suggestion
await tester.pumpAndSettle();
```

TheTexanCodeur (commented):
Great work!

This new circular value implements an efficient and elegant way of handling errors. This new tool will be really helpful in the future. 

I just have one suggestion but overall LGTM!

test/component/utils/ui/circular_value_test.dart
Aderfish: Thank you for the suggestion, I added it.

camillelnne (approved):
LGTM!
The code is super clean :)

I was able to test it on a physical device, works perfect

TheTexanCodeur (approved):
LGTM!

PR #128, opened 2024-04-17
Geoflutterfireplus refactor

yoannLafore:
Hello,
This PR aims to provide a proper fix to #78.
Indeed, a first fix was proposed in PR #116. 
But this implied that the overview was displaying a stream that was auto updated in real time.
After further discussions with the team, we determined that this wouldn't contribute to the user experience.
For instance, if the user is reading a post and that another is added near him, his scroll position might be offseted which can be anoying.
Thus, we chose to stay with a manual refresh for now.

In order to have a manual refresh, we need to be able to expose a `Future<List<PostFirestore>>` in `PostRepository` as previously while correcting the bug.
Technically speaking, this requires the underlying library to provide a `Future` implementation of geo queries. This feature is currently not supported by `GeoFlutterFire2`.

After some extensive researches, I found out that there exists a newer version of this library called **GeoFlutterFire plus**.
This library seems to be cleaner, maintained with thorough testing and it provides the functionnality that we described above.

Thus, in this PR, we do :
- Refactor the codebase to use **GeoFlutterFire plus**
- Use the `Future` implementation of **GeoFlutterFire plus** to get the nearby posts which fixes the bug #78
- Re-enable the firestore cache which was only a temporary fix provided in #79.

## Live testing
You can live test this fix as follows:
- Go into the overview page
- Add a post through the application
- Refresh the overview

With the presence of the bug, the new post wouldn't appear, but with this fix, you should be able to see it appear in the feed.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1,4 +1,3 @@
-import "package:cloud_firestore/cloud_firestore.dart";
 import "package:firebase_core/firebase_core.dart";
 import "package:flutter/material.dart";
 import "package:flutter/services.dart";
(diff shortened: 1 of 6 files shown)


gruvw (approved):
Awesome finding! `geoflutterfire_plus` looks way more promising than `geoflutterfire2`. Very clear and comprehensible changes. Good to go :)

yoannLafore:
Thanks you for your feedback !

PR #144, opened 2024-04-21
Refactor tests

### LINKED ISSUE #131: Refactor test package structure and providers

camillelnne:
Fixes #131.

This PR introduces a significant refactoring of our test suite aimed at standardizing testing processes and enhancing maintainability. The goal was to eliminate inconsistencies, improve the accessibility of mocks, and align the test directory structure more closely with our lib directory to minimize confusion. 

**Key Changes:**
- **Test Folder Refactoring**: Reorganized the test folder structure to mirror the `lib `directory more closely, to avoid confusion when creating new tests.
- **Centralized Mock Data and Services**: Introduced a new mock directory that consolidates all our fake data, mock services, and provider configurations. 
- **Uniform Setup in View Tests**: Standardized the setup across all view tests to ensure consistency in how tests are initialized and run. 
- **Centralized Provider Scopes**: Moved all `ProviderScope` widgets to a centralized location. This decision should be discussed with the reviewers (see later paragraph).

We now have a big toolbox of mocks, that could be reused in future tests :)

**Discussion points**:
**Centralized Provider Scopes**: In my second to last commit "abstract away providers", I have moved all providers to dedicated files, as variables or methods to be called in the views test files. However, it occurred to me that some providers parameters may change in the future, e.g the page now needing a mocked user repository or a mock geolocation service. Therefore, I'd like to have the opinion of reviewers: should we keep this centralization or not?  (if not, I will simply revert the commit.)
For better understanding, please have a look at the files in `mock/providers` to see the implementation and differences between each, and check the diff in commit ""abstract away providers"  (SHA: `3220db20b7e696bf38823896789559680072eea2`).

**Acceptance criteria**:
- Change the test packages to a much cleaner one (see my review of PR https://github.com/ProximaEPFL/proxima/pull/122). So, have a package model test, a package for viewmodel test, a package for ui test, and mocks.
- The mock package, there all mocks, mock data and overrides in list.
- Decrease the amount of data they create, centralising it in the mocks (if someone creates a fake user, just centralise it in mocks so that everyone uses the same)
- Make all tests use overrides defined in mocks.


gruvw:
It's a detail, just so you know for next time, those commits should probably have been tagged by the `refactor:` tag from conventional commits instead of `fix:` (which is more for bug fixes).

test/models/database/post/post_data_test.dart
```diff
@@ -6,6 +6,7 @@ import "package:proxima/models/database/user/user_id_firestore.dart";
 void main() {
   group("Post Data testing", () {
     test("hash overrides correctly", () {
+      //TODO: moves this mock to package
```
CHOOSEIT, line 9: I think that we can move this mock now.

I think it also duplicates with `test/models/database/post/post_firestore_test.dart` line 20

+ little typo: `post_ti**l**tle`

test/services/database/post_repository_test.dart
```diff
@@ -48,22 +46,12 @@ void main() {
...
-        ownerId: const UserIdFirestore(value: "owner_id"),
-        title: "post_tiltle",
-        description: "description",
-        publicationTime: Timestamp.fromMillisecondsSinceEpoch(4564654),
-        voteScore: 12,
-      ),
+    final post = MockPostFirestore.createPostAt(
+      MockPostFirestore.generatePostData(1).first,
+      userPosition2,
     );
```
CHOOSEIT, line 52: I think that we can maybe abstract this method, as I saw duplication of post creation at a certain location without using the actual post generated.

So we may want a function that given a location create a post, without needing to provide `PostData`. This function could be named: `generatePostAt` in the `MockPostFirestore`

Alternative representation (line 90, 80, 104, 118): `final postData = MockPostFirestore.generatePostData(1).first;
      final expectedPost =
          MockPostFirestore.createPostAt(postData, farAwayPostPosition);`

test/mocks/data/mock_position.dart
```diff
@@ -0,0 +1,49 @@
...
+  final postsNotInRange = List.generate(outRange, (i) {
+    i = i + 7; // makes it out of range
+    return GeoPoint(
+      userLatitude + 0.0001 + i * 0.0001,
+      userLongitude + 0.0001 + i * 0.0001,
+    );
+  });
+
+  return [...postInRange, ...postsNotInRange];
+}
```
CHOOSEIT, line 49:
```suggestion
/// Generate a list of [GeoPoint] positions, in and out of range
List<GeoPoint> generatePositions(
  GeoPoint userPosition,
  int inRange,
  int outRange,
) {
  assert(inRange >= 0 && outRange >= 0);

  const maxDiagonalDistanceInRange = 0.0005;
  const minDiagonalOutRange = 0.001;
  double distanceBetweenPosts = maxDiagonalDistanceInRange / inRange;

  double userLatitude = userPosition.latitude;
  double userLongitude = userPosition.longitude;

  final postInRange = List.generate(inRange, (i) {
    double dDirection = distanceBetweenPosts * i;
    return GeoPoint(
      userLatitude + dDirection,
      userLongitude + dDirection,
    );
  });

  // Generate post positions that are not in the range.
  final postsNotInRange = List.generate(outRange, (i) {
    double dDirection = minDiagonalOutRange + distanceBetweenPosts * i;
    return GeoPoint(
      userLatitude + dDirection,
      userLongitude + dDirection,
    );
  });

  return [...postInRange, ...postsNotInRange];
}
```

I think that we could create a function that accept any number of posts in InRange

CHOOSEIT (changes requested):
Overall looks good to me.

Huge and clean work 💪🏼 

We still have to discuss about the `mocks/providers` folder.

I think that for the moment we cannot know what we will need in a few sprints, so we may just add our provider and use the one you created as template.

Thank you !

test/mocks/data/mock_position.dart
camillelnne: Hey! The method I have created actually accepts any number of posts inRange, just forgot to update the documentation. 
Yours looks cleaner though, will test and commit it, thanks!

CHOOSEIT: I divided the max diagonal distance by InRange 

As i goes from 0 to InRange, the last element will be at the maximal diagonal distance

test/mocks/data/mock_position.dart
```diff
@@ -0,0 +1,49 @@
...
+const nearbyPostPosition = GeoPoint(40.0001, 20.0001); // 14m away
+const farAwayPostPosition = GeoPoint(40.001, 20.001); // about 140m away
+const postOnEdgeInsidePosition = GeoPoint(
+  39.999999993872564,
+  20.001188563379976 - 1e-5,
+); // just below 100m away
+const postOnEdgeOutsidePosition = GeoPoint(
+  39.999999993872564,
+  20.001188563379976 + 1e-5,
+); // just above 100m away
```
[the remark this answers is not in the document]
camillelnne: What do you think of creating methods instead? So that we can get positions relative to any GeoPoint.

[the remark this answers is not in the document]
CHOOSEIT: That is a great idea !

test/mocks/services/mock_user_repository_service.dart
yoannLafore, line 1: I am not sure of moving `_mockEmptyFirestoreUser` to `test/mocks/data/mock_firestore_user.dart/mockEmptyFirestoreUser` and making it public.

The idea behind this mock was simply to be here to satisfy the `Mockito` convention in the override method. But it is actually never accessed anywhere because the value will be overriden when we'll do
```dart
when(userRepo.getUser(***).thenAnswer((_) => ***))
```
in the tests.
Thus we never actually use it in the tests.
I believe that we might want to keep it private in that file.
What do you think ?

test/models/database/post/post_data_test.dart
```diff
@@ -23,8 +18,14 @@ void main() {
       );
 
       final actualHash = data.hashCode;
-
       expect(actualHash, expectedHash);
+
+      const geoPoint = GeoPoint(40, 20);
```
yoannLafore, line 23: I believe we might use `test/mocks/data/mock_position.dart/userPosition1` here.
```suggestion
      const geoPoint = userPosition1;
```

test/mocks/overrides/override_auth_providers.dart
```diff
@@ -5,7 +5,7 @@ import "package:google_sign_in_mocks/google_sign_in_mocks.dart";
 import "package:hooks_riverpod/hooks_riverpod.dart";
 import "package:proxima/services/login_service.dart";
 
-import "../test_data/firebase_auth_user_mock.dart";
+import "../data/mock_firebase_auth_user.dart";
 
 final firebaseAuthMocksOverrides = [
```
JoachimFavre, line 10: I don't think this is necessary for all providers, but I'm feeling this authentication providers need some small documentation to be very clear:
```suggestion
/// Override to use to test the login process
final firebaseAuthMocksOverrides = [
```

test/mocks/providers/provider_login_page.dart
```diff
@@ -0,0 +1,26 @@
...
+import "../overrides/override_home_view_model.dart";
+
+ProviderScope loginPageProvider(FakeFirebaseFirestore fakeFireStore) {
+  final userRepo = UserRepositoryService(firestore: fakeFireStore);
+
+  return ProviderScope(
+    overrides: [
+      ...firebaseAuthMocksOverrides,
+      ...mockEmptyHomeViewModelOverride,
+      userRepositoryProvider.overrideWithValue(userRepo),
```
JoachimFavre, line 18: By humble curiosity, how did you choose which overrides to extract in their own package, and which to keep in the providers? I have seen multiple overrides in various the providers.

test/models/database/post/post_data_test.dart
JoachimFavre: Well seen!
If I may add, I think it may also apply to all `GeoPoint(40, 20)` of `post_upvote_repository_test.dart` and `post_repository_test.dart` too (if you agree with my reasoning, probably Ctrl+F the whole test folder for `GeoPoint(40, 20)` and `GeoPoint(0, 0)`)

test/views/pages/home/navigation_home_test.dart
```diff
@@ -1,54 +1,29 @@
...
-          () => MockHomeViewModel(
-            build: () async => testPosts,
-          ),
-        ),
-      ],
-      child: homePageApp,
-    );
+    emptyHomePageWidget = emptyHomePageProvider;
+    nonEmptyHomePageWidget = nonEmptyHomePageProvider;
   });
```
JoachimFavre, line 20: By humble curiosity, why did you choose to do this this way and not simply replacing `emptyHomePageWidget` and `nonEmptyPageWidget` by `emptyHomePageProvider` and `nonEmptyHomePageProvider`? Similarly, why not simply doing
```
ProviderScope emptyHomePageWidget = emptyHomePageProvider;
ProviderScope nonEmptyHomePageWidget = nonEmptyHomePageProvider;
```

Similar question for most of tests in the `pages` package (when the provider does not depend on some parameter).

JoachimFavre (commented):
This is an incredible work. The tests now go straight to the point, I love how clear they are. I think the diff of `circular_value_test` shows a summary of the importance of your work, good job!

I just have some quick comments you can find below, and some others, more general, here.

First, I personally think `post_upvote_repository_test` should use the mock data defined in `mocks`. 

Second, I'm not sure I get the difference between `models` and `services`. For instance, why is `post_upvote_repository` a service? I think it was @yoannLafore who made this test structure originally, so maybe he is part of this discussion too. Personally, I think we should consider merging the packages `models` and `services` (effectively putting the content of `services` inside `models`).

Third, there are some `GeoPoint` inside `home_view_model_integration_test` which I personally think should be replaced `nearbyPostPosition` and `farAwayPostPosition`.

Again, thank you so much for your work!

JoachimFavre:
> Centralized Provider Scopes: In my second to last commit "abstract away providers", I have moved all providers to dedicated files, as variables or methods to be called in the views test files. However, it occurred to me that some providers parameters may change in the future, e.g the page now needing a mocked user repository or a mock geolocation service. Therefore, I'd like to have the opinion of reviewers: should we keep this centralization or not? (if not, I will simply revert the commit.)
> For better understanding, please have a look at the files in mock/providers to see the implementation and differences between each, and check the diff in commit ""abstract away providers" (SHA: 3220db20b7e696bf38823896789559680072eea2).

I personally really like what you did. I cited `test/utils/ui/circular_value_test.dart` in my review comment, and I think this a great example. With this abstraction, the tests become so much clearer. 

I'm however not sure I completely understood your issue. If suddenly a page needs a mock geolocation service, then all tests that want to test this page will need to also add this override, so the abstraction is useful; isn't it? If you think that we might just get many providers in the long term and that the package will just have many files, I think this an issue we can leave for later; it should not be too hard to fix (by adding sub-packages) with the great work you have done on the test structure.

test/models/database/post/post_data_test.dart
camillelnne: Yes, that was definitely the point of creating userPosition1, thanks for catching it!

test/mocks/providers/provider_login_page.dart
camillelnne: Good question!
I actually had a conversation with Yoann on our group about it.
It is because we reuse the parameters.

Attaching the screenshot :)
image

test/views/pages/home/navigation_home_test.dart
camillelnne: You mean defining it in the setup instead of doing it once at the beginning ?

yoannLafore:
> Second, I'm not sure I get the difference between models and services. For instance, why is post_upvote_repository a service? I think it was @yoannLafore who made this test structure originally, so maybe he is part of this discussion too. Personally, I think we should consider merging the packages models and services (effectively putting the content of services inside models).

I think the difference between a service and a model is the following:
- The models represent the data that our application uses (the firestore dataclasses, etc.)
- The services are here to expose the model data to higher levels. Thus, `PostRepositoryService` will expose the `PostFirestore` to the higher levels of the application (i.e. the viewmodels)

I believe it is important to keep the distinction between models and services to keep the codebase well-separated.

test/mocks/data/mock_firestore_user.dart
```diff
@@ -4,6 +4,34 @@ import "package:proxima/models/database/user/user_data.dart";
...
+
+final testingUserFirestoreId = UserIdFirestore(value: testingLoginUser.uid);
+
+final testingUserFirestore = UserFirestore(
+  uid: testingUserFirestoreId,
+  data: testingUserData,
+);
+
 /// Helper class to generate mock user data to be used in tests
 class MockUserFirestore {
```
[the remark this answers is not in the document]
camillelnne: Most of those names are names I kept from the previous implementation, but you're right it makes more sense!

test/mocks/services/mock_user_repository_service.dart
camillelnne: Hi, I actually used it in one test, after @JoachimFavre suggestion. Let's discuss this further next standup!

yoannLafore: Yes, good idea !

gruvw:
Is there a reason why this PR reduces the coverage ? I thought it was only refactoring. It's not a big problem, as we are still well over the 80% required, but I was just wondering out of curiosity why we lose 1 percentage point in a refactoring process.

test/mocks/services/mock_user_repository_service.dart
camillelnne: Resolved in commit `6587023ce03873a49422a3b961b4d26626e9908c`

test/mocks/providers/provider_login_page.dart
```diff
@@ -0,0 +1,26 @@
...
+import "package:hooks_riverpod/hooks_riverpod.dart";
+import "package:proxima/services/database/user_repository_service.dart";
+import "package:proxima/views/navigation/routes.dart";
+import "package:proxima/views/pages/login/login_page.dart";
+
+import "../overrides/override_auth_providers.dart";
+import "../overrides/override_home_view_model.dart";
+
+ProviderScope loginPageProvider(FakeFirebaseFirestore fakeFireStore) {
+  final userRepo = UserRepositoryService(firestore: fakeFireStore);
```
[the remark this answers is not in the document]
yoannLafore: After discussion in person, we determined that this provider is only used to allow the pumping of the UI widgets.
Thus, having access to the user repository is not necessary and we can let the code as it is.

test/mocks/providers/provider_profile_page.dart
```diff
@@ -0,0 +1,27 @@
...
+import "package:proxima/services/login_service.dart";
+import "package:proxima/views/navigation/routes.dart";
+import "package:proxima/views/pages/profile/profile_page.dart";
+
+import "../overrides/override_auth_providers.dart";
+
+ProviderScope profilePageProvider(FakeFirebaseFirestore fakeFireStore) {
+  UserRepositoryService userRepo = UserRepositoryService(
+    firestore: fakeFireStore,
+  );
```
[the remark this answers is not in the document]
yoannLafore: We did not choose to apply the suggestion (see comment in above suggestion).

yoannLafore (approved):
Thank you for your amazing work :tada !
The PR looks good to me :+1: 
This will really help us to have nice and consistent tests.

Great job :tada:

camillelnne:
> Is there a reason why this PR reduces the coverage ? I thought it was only refactoring. It's not a big problem, as we are still well over the 80% required, but I was just wondering out of curiosity why we lose 1 percentage point in a refactoring process.

Hey, good catch @gruvw!
Found the reason why: in commit 942a0a230fdd918fb566bf81fe6baa62f883c062, for the login page test, the I changed the home of the `ProviderScope` from `ProximaApp` to `LoginPage`, which made more sense for a unit test in my opinion.
Thus we are now missing coverage on the `ProximaApp`, but this should be fixed by the end to end tests, as @yoannLafore was able to integrate them in the CI recently.

CHOOSEIT (approved):
LGTM ! great job thank you !

test/views/pages/home/navigation_home_test.dart
JoachimFavre: In most of the tests in the `pages` package, you do something like:

```dart
late ProviderScope xWidget;

setUp(() async {
   xWidget = xProvider;
});

testWidget("", (tester) async {
     await tester.pumpWidget(xWidget);
});
```

Why not replacing this by:
```dart
testWidget("", (tester) async {
     await tester.pumpWidget(xProvider);
});
```

Or, if you want to keep the modularity of an external variable (which I personally don't think is necessary):
```dart
ProviderScope xWidget = xProvider;

testWidget("", (tester) async {
     await tester.pumpWidget(xWidget);
});
```

test/mocks/data/mock_firestore_user.dart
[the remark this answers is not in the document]
JoachimFavre: Thank you so much for fixing this. Browsing through the files, I'm just realising that the file names may make sense to change too. 
For instance, `mock_firestore_post.dart` to `firestore_post_generator.dart`, `mock_firestore_user.dart` to `firestore_user_generator.dart` and `mock_post_data.dart` to  `post_data_generator.dart`.

JoachimFavre (approved):
Except for some small discussions that still need to be resolved, this LGTM.

This is not very important in this case, but I found that some commits of this PR were quite hard to review because they had many unrelated things together. Among others, I am thinking of the number of things done in aaa5d488755fff71ea1e669d9b539ef4b1427aa2 or, less importantly,  the `GeoPoint(0, 0)` getting replaced by `userPosition0` in e2911c5a43b8c751f9ce204c5627df7e082d631b. 
Note that doing small commits was harder than usual in this PR since there was many things that needed to be done, and that we therefore want to do everything at once when we get the idea. I just think this is interesting to point out for self-improvement! :)

test/views/pages/home/navigation_home_test.dart
camillelnne: I see what you mean.
To be fair, it was the way most tests were written (all except for the circular value it seems), by putting the provider in the setup, so I kept this convention.

JoachimFavre: Wasn't this because creating a provider was lot of code? Now that they are already defined somewhere else, is it really necessary to keep those variables?

test/mocks/data/mock_firestore_user.dart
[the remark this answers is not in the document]
camillelnne: Done in e1403dc838da1a2512cd56075d2d892f85382e06 :)

test/models/database/post/post_data_test.dart
camillelnne: Oh thanks! I remember doing a ctrl+f to find those, but somehow I missed it.
Fixed in 2b8946e972325a8a7222a3408678243b03ff332d

test/views/pages/home/navigation_home_test.dart
camillelnne: fixed in 99810edc31106c3d02c2945ae2fe730ba6a58bc6.
I didn't replace the providers that took a parameter, for abstraction purpose.

Issue #135, opened 2024-04-19
Make viewmodel for upvotes


JoachimFavre:
Acceptance criteria:
- See if a user upvoted/downvoted a post
- Let users change their vote on a post
- The database read must be minimal
- When the user upvote/downvote a post, the result (up/down arrrow highlighted) must be visible immediately without a manual refresh

PR #145, opened 2024-04-23
Votes viewmodel

yoannLafore:
Hello,

This PR aims to provide the view model to allow the user to vote on a post (#135).
In particular, this PR implement the following:

#### Key implementations:

- **Dataclass for the vote UI:** This PR adds a new dataclass `PostVote` that corresponds to the UI data for the votes of a post. It contains `votes` that represents the number of votes of the post. It also contains `upvoteState` which is the current state of the user vote (upvote, downvote,  none)

- **View model for the votes**: This PR adds the `UpVoteViewModel`. This viewmodel takes as parameter the `PostIdFirestore` of its corresponding post. It exposes exposes the `PostVote` state for the post for which it is responsible.
It provides the functions `triggerUpvote` and `triggerDownvote` to allow the user to perform vote actions.

- **Expose post ids in the home view model:** The new `UpVoteViewModel` requires to access the firestore document of a post. Thus, it needs to know the `PostIdFirestore` of the post. This is why we expose the post ids in the `HomeViewModel`.

- **Mocks:** This PR provides a Mock abstraction of the `PostUpvoteRepositoryService`.

- **Tests:** This PR provides unit and integration tests for the view model.

#### Details about `UpVoteViewModel`:
To reduce the number of fetches from the database, the view model implements a local approach for exposing the state modifications. Indeed, the initial vote state and number of votes are queried once during the `build()` of the view model.
Then, when a `triggerUpvote` or `triggerDownvote` is called, the vote is registered on the database but the state is refreshed with local data (we do not query again the number of votes from the database). This allows to avoid a database fetch while giving a responsive experience to the user.

If the `triggerUpvote` is called and that the vote state was alrealy an upvote, the upvote will be removed. This allows the user to remove an upvote on a post. The same applies for downvotes.

#### Live testing:
You can live test the voting logic by going in the overview and voting on some posts. You should see your vote being highlighted as well as the vote counter being updated accordingly.

#### Acceptance criteria:
- See if a user upvoted/downvoted a post
- Let users change their vote on a post
- The database read must be minimal
- When the user upvote/downvote a post, the result (up/down arrrow highlighted) must be visible immediately without a manual refresh

**Note:** Waiting for #144 to be merged.


lib/viewmodels/upvote_view_model.dart
```diff
@@ -0,0 +1,97 @@
...
+      ),
+    );
+
+    await voteRepository.setUpvoteState(uid, postId, newVoteState);
+  }
+}
+
+final postVoteProvider = AsyncNotifierProvider.autoDispose
+    .family<UpVoteViewModel, PostVote, ({PostIdFirestore postId})>(
+  UpVoteViewModel.new,
```
gruvw, line 96: Here the code should not use this `.new` syntax as it was an old Dart way of creating objects that is deprecated today.

lib/viewmodels/upvote_view_model.dart
```diff
@@ -0,0 +1,97 @@
...
+///
+/// To avoid unnecessary database fetches, the [PostVote] state is queried
+/// from the database only once at build time and then updated locally after the
+/// user triggers an upvote or downvote action.
+///
+/// This view model must be AutoDisposed because it is created on the fly
+/// for each post that is displayed on the screen.
+/// Thus, when the post is removed from the screen, the view model is disposed.
+class UpVoteViewModel extends AutoDisposeFamilyAsyncNotifier<PostVote,
+    ({PostIdFirestore postId})> {
```
gruvw, line 26: We should replace this type tuple by the actual `PostIdFirestore` type of its own. It does not really matter if it needs to be called `arg` on the overridden method.

lib/viewmodels/home_view_model.dart
```diff
@@ -1,18 +1,23 @@
...
 import "package:proxima/services/geolocation_service.dart";
 
 /// This viewmodel is used to fetch the list of posts that are displayed in the home feed.
-/// It fetches the posts from the database and returns a list of [PostOverview] objects to be displayed.
-class HomeViewModel extends AsyncNotifier<List<PostOverview>> {
+/// It fetches the posts from the database and returns a list of
+/// (postId: [PostIdFirestore], postOverview: [PostOverview]) objects to be displayed.
+/// These represent the overview data to be displayed associated to the corresponding post id.
+class HomeViewModel extends AsyncNotifier<
+    List<({PostIdFirestore postId, PostOverview postOverview})>> {
```
gruvw, line 13: Also here this code should not make use of a tuple type as it hinders readably and usability of this type.
Why not simply put the `PostIdFirestore postId` inside the `PostOverview` type as a new field ?
I think it would make much more sense that way.

gruvw: This change would also simplify the map in the `HomeViewModel.build` method.

lib/views/home_content/feed/post_card/post_card.dart
```diff
@@ -30,7 +33,10 @@ class PostCard extends StatelessWidget {
       child: Row(
```
gruvw, line 33: Here I think we could just order the local variables as they would appear in the view, so placing the `postBottomBar` at the end of the build would facilitate code reading.

lib/views/home_content/feed/post_card/votes_widget.dart
```diff
@@ -1,43 +1,59 @@
...
     super.key,
-    required this.votes,
+    required this.postId,
   });
 
   @override
-  Widget build(BuildContext context) {
+  Widget build(BuildContext context, WidgetRef ref) {
+    final asyncPostVote = ref.watch(postVoteProvider((postId: postId)));
+    final votes = asyncPostVote.valueOrNull?.votes ?? 0;
```
gruvw, line 20: Interesting choice here to use `valueOrNull` instead of a circular value...
Are we sure this is the behavior we want ?

I don't have a definite answer as this is a case where it would maybe not be the best UX to have spinners on every vote count till the request resolves but in the other hand we have no error handling and this differs from our standard way to make async requests to DB in other parts of the project.

lib/views/home_content/feed/post_card/votes_widget.dart
```diff
@@ -1,43 +1,59 @@
...
+    final upvoteState =
+        asyncPostVote.valueOrNull?.upvoteState ?? UpvoteState.none;
+
+    final upvoteNotifier =
```
gruvw, line 24: Please stack all `ref.` accesses to the top of the build method as they can introduce side effects and are quite tricky to track otherwise.

lib/views/home_content/feed/post_card/votes_widget.dart
```diff
@@ -1,43 +1,59 @@
...
+        size: 30,
+        color: upvoteState == UpvoteState.upvoted ? Colors.blue : null,
+      ),
       style: const ButtonStyle(
         tapTargetSize: MaterialTapTargetSize.shrinkWrap,
       ),
-      onPressed: () {
-        //TODO: Implement the logic to upvote the post
-      },
+      onPressed: () => upvoteNotifier.triggerUpVote(),
```
gruvw, line 40: You can use shorthand notation here `onPressed: upvoteNotifier.triggerUpVote` try to apply this advice wherever possible.

lib/views/home_content/feed/post_card/votes_widget.dart
```diff
@@ -1,43 +1,59 @@
...
     );
 
     final downvote = IconButton(
       //Used to reduce the padding and InkWell created by the IconButton widget
```
gruvw, line 44: Avoid duplicating comments as it makes them harder to maintain.

gruvw:
Awesome work on this feature! Also, thank you for the live demo earlier. Just made a few comments and other than that, this code looks good to be merged to me :rocket:

test/viewmodels/upvote_view_moodel_integration_test.dart
CHOOSEIT, line 1: Little typo:
test/viewmodels/upvote_view_m**oo**del_integration_test.dart

lib/viewmodels/home_view_model.dart
yoannLafore: My initial idea was that `PostOverview` represents only the data that are displayed on the screen.
Thus, adding a `PostIdFirestore` to it would break this abstraction, that's why I chose to return a list of tuples instead.
But effectively, if you think that this will contribute the code readibility, I can add the `postId` into the `PostOverview` directly.

lib/viewmodels/upvote_view_model.dart
yoannLafore: Oh, I didn't know this was a deprecated feature.
Thanks for pointing it out !
Fixed in 5b00f1abc10da9c397cb53ac476a1681227abf8a.

lib/viewmodels/upvote_view_model.dart
yoannLafore: I had originally done that because I though it would made the code easier to understand.
But looking back at it, this adds quite some complexity and I believe that this doesn't really worth it actually. 
So thank you for pointing it out :+1:

lib/views/home_content/feed/post_card/post_card.dart
yoannLafore: Thanks for the advice !
Fixed in e6c59a96bc835188c350722a647fef6e163bee69.

lib/views/home_content/feed/post_card/votes_widget.dart
yoannLafore: Good advice !
I do that in b43e73420a4775989b68de80142e6329265da0d7.

lib/views/home_content/feed/post_card/votes_widget.dart
yoannLafore: I didn't know this notatiuon existed.
Thanks for bringing it to me !

I make use of it in 10a282429160146dc9f5da558bce2336eb176f5f.

test/viewmodels/upvote_view_moodel_integration_test.dart
yoannLafore: Thanks for pointing it out !
Fixed in f915d73c957e390828fb8ae22eff908cd12f5d55.

lib/viewmodels/home_view_model.dart
gruvw: Yeah its not only for displayed elements, it's everything the view layer might need.

lib/viewmodels/upvote_view_model.dart
gruvw: Yes, much better thx.

gruvw:
As discussed during the standup meeting today, if you fixed a straight forward comment review, don't hesitate to mark it as "Resolve conversation" right away after including the commit hash so we don't spend more time on it :)

lib/views/home_content/feed/post_card/votes_widget.dart
yoannLafore: You are right, I remove the duplication in 63fecb17c93cd8cb541334a9ecd53d0d851ba7bc.

lib/views/home_content/feed/post_card/votes_widget.dart
yoannLafore: To answer @gruvw question, I believe indeed that this is the behavior that we want.
In our application, we used a circular value at the places where there is only one element to wait for.
For instance, we have one circular value for the whole list of post.

Here, as you said we would need to have a circular value for each post. I believe this can be quite distracting for the user. Thus I believe that it is better to simply display "nothing" (no vote and 0 as number of votes) in the loading state.
Regarding error handling, I think it is better to silently ignore the errors as it allows the user to continue using the app without this feature (which isn't a absolute necessity to use the app).

If we want to provide an even smoother transition from the loading state to the data, I guess we could pass the initial number of votes as parameter to the widget (coming from the corresponding `PostOverview`). And we display this number of votes while the `UpVoteViewModel` is loading. But this is only an idea, what do you think ?

yoannLafore: To answer @CHOOSEIT question, we cannot use the `voteScore` coming from the `postOverview` to instanciate the `UpVoteViewModel`.
I also had this idea originaly but this leads to the following bug:

To limit ressources and db calls, the `UpVoteViewModel` are created on the fly only when the corresponding post is displayed on the screen. 
Thus, it can be created for two reasons:
- (1) The user manually refreshes the overview which rebuild the widget and so the view model
- (2) The user scroll down (thus the viewmodel is disposed) and then up again (thus the viewmodel is recreated)

These two possibilities implies that we cannot know for sure that the `voteScore` used to create the viewmodel is coherent with the database.
Indeed, consider the following scenario:
- The user opens the overview and see "postA" which has 10 votes.
- The user upvotes "postA".
- Thus, the viewmodel exposes locally a vote score of 11 but the `voteScore` from `postOverview` is still 10.
- The user scrolls down thus the viewmodel is disposed.
- Thue user scrolls back up .
- The viewmodel is re-instanciated with the `voteScore` of the `postOverview` which is 10 and it exposes this value.
- Thus we have an incoherent number of votes displayed.

This is why we need query the number of votes in the viewmodel. 
With that, we only need the `postId` as widget parameter.

yoannLafore:
Thank you for your great feedback :tada: 
I have try to fullfilled as much as possible your suggestions.

Regarding the question of @CHOOSEIT  on UI testing.
I think that this is indeed a good idea and after rapidly trying out it seems to be possible.

However, in order to keep a clean test architecture, this would requires to add quite a few mocks/overrides.
I think that this would make this PR (that is already quite big) too big.
Thus I believe we should add such tests in another PR.
But this is a very good idea, thanks for pointing it out :+1: 

Also, thanks for the suggestion @gruvw, in the future I will to resolve conversations whenever there are no ambiguities.

gruvw:
I agree with the decision of placing the UI testing in another PR.

CHOOSEIT (approved):
Looks good to me ! 

Great job thank you !

CHOOSEIT (changes requested):
After some additional little testing, I noticed that when I disable the internet connection and I try to upvote / downvote, the application crashes I don't know if we should handle this behavior in this PR.

As we are not using the circularValue here, we don't have the Albert's error handling

lib/views/home_content/feed/post_card/votes_widget.dart
gruvw: Maybe we could not have it as auto-dispose ?

yoannLafore: The issue if we not have it as auto-dispose is that the viewmodel won't be recreate when the user refreshes manually.
Thus, the number of votes wont refresh.

yoannLafore:
Good catch @CHOOSEIT, thank you for the remark ! 
Effectively, the exception is not very well managed.
I don't exactly know what approch we want to take regarding those kinds of errors however.
Thus I propose that we move this into the product backlog to be fix into another PR.

gruvw:
The proper error handling decision can be moved to the PB, but I think we should not have the app crashing in that case...
I think if we could just ignore the error and still have the app be useable (after the error happens) before merging, it would be better.

yoannLafore:
Currently, when the app is launched from the phone directly, the error is ignored and the app is still usable.
I can only see the exception being thrown when vscode is monitoring which pauses the execution.
However, after continuing the execution, the app doesn't crash and is still usable (but the number of votes displayed is incorrect).
Thus I believe it can stay like that until we fix #160.

gruvw (approved):

CHOOSEIT (approved):
Looks good to me ! Great job, thank you ! 🎉

PR #149, opened 2024-04-24
Add a post exists method

Aderfish:
Just a small PR to add a postExists method to the post repository service. The method returns whether a post id exists in the database. I found I needed it for the challenge service.


test/services/database/post_repository_test.dart
```diff
@@ -233,5 +233,17 @@ void main() {
...
+      await setPostFirestore(post);
+
+      final exists = await postRepository.postExists(post.id);
+      expect(exists, true);
+
+      await postRepository.deletePost(post.id);
+
+      final existAfterDeletion = await postRepository.postExists(post.id);
+      expect(existAfterDeletion, false);
+    });
```
JoachimFavre, line 247: I think verifying the post does not exist before its creation adds some useful robustness to the test (this allows to test a post does not exist when it has been deleted, but also when it has never existed). I therefore propose:
```suggestion
    test("post exists works", () async {
      final existsBeforeCreation = await postRepository.postExists(post.id);
      expect(existsBeforeCreation, false);

      await setPostFirestore(post);

      final exists = await postRepository.postExists(post.id);
      expect(exists, true);

      await postRepository.deletePost(post.id);

      final existAfterDeletion = await postRepository.postExists(post.id);
      expect(existAfterDeletion, false);
    });
```

JoachimFavre (approved):
Just a small comment above, but otherwise, LGTM!

JoachimFavre (approved):
I just have a small comment below, but otherwise, LGTM!

EDIT: Whoops, it appears I made a small mistake with VSCode's GitHub extension and reviewed twice bahaha.

yoannLafore (approved):
Thanks for bringing this new method !
The PR looks good to me :+1:

PR #151, opened 2024-04-24
Fix tests

### LINKED ISSUE #156: Fix tests
Acceptance criteria: fix the following tests:
- `post_repository_test.dart:"Get simple user posts correctly"`, which fails consistently on some machines, but not on others.
- `new_post_test.dart:"Writes non empty post to repository"`, which fails in a non-deterministic fashion.

JoachimFavre:
Fixes #156.

This PR fixes some tests, which failed in a random-looking fashion.
1) 1d58e3ac1fbe6d507d46e61133e38a6b6b65f2bf's fix is thanks to @yoannLafore, who found the problem. Two posts could have the same id, and thus replace one another in the database. This failed consistently on some computers, but not on others.
2) c6ed8db2916fa65a48d585a89bca9f2897b9e313's fix simply increases the allowed error margin. It was previously too small, resulting in a rare, but non-deterministic, fail.

While reviewing, I want to particularly attract your attention to 9c46f5492a92c962bad80d9fe05be11e3373b7a9. This is done just to uniformise with the first fix. This may not have its place in this PR, and it may just be complicating everything. Don't hesitate to tell me if you want me to revert it.

### DIFF
--- a/test/services/database/post_repository_test.dart
+++ b/test/services/database/post_repository_test.dart
@@ -109,7 +109,9 @@ void main() {
       const userPosition = userPosition1;
       final expectedPost = FirestorePostGenerator().generatePostAt(
         geoPointGenerator.createPostOnEdgeInsidePosition(
-            userPosition, kmRadius),
+          userPosition,
+          kmRadius,
+        ),
       );
 
       await setPostFirestore(expectedPost);
@@ -123,7 +125,9 @@ void main() {
       const userPosition = userPosition1;
       final expectedPost = FirestorePostGenerator().generatePostAt(
         geoPointGenerator.createPostOnEdgeOutsidePosition(
-            userPosition, kmRadius),
+          userPosition,
+          kmRadius,
+        ),
       );
 
       await setPostFirestore(expectedPost);
--- a/test/services/database/post_repository_test.dart
+++ b/test/services/database/post_repository_test.dart
@@ -193,20 +193,17 @@ void main() {
     });
 
     test("Get simple user posts correctly", () async {
+      final generator = FirestorePostGenerator();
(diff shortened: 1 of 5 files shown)


test/services/database/post_repository_test.dart
```diff
@@ -119,7 +119,7 @@ void main() {
 
     test("Post on edge (outside) is not queried", () async {
       const userPosition = userPosition1;
-      final expectedPost = FirestorePostGenerator.generatePostAt(
+      final expectedPost = FirestorePostGenerator().generatePostAt(
```
gruvw, line 122: Shouldn't we have just one instance per test ? Why recreate an instance every time we call `generatePostAt` ?

gruvw (approved):
Made a minor change and apart from my comment this LGTM :)

test/services/database/post_repository_test.dart
gruvw: We could maybe use `setUp` function with late assignments like we do with mocks to ensure a clean instance is used for every test ?

TheTexanCodeur: I agree with this comment. @gruvw 's proposal will make this code respect our new test standards.

TheTexanCodeur (approved):
LGTM!

Good job finding and resolving those issues. I agree with @gruvw 's suggestion, but overall this PR is ready to merge in my opinion.

test/services/database/post_repository_test.dart
JoachimFavre: My original idea was that this would be adding unnecessary boilerplate. I however understand your point of view, and apply the majority's opinion! :)
Fixed in db6722bd9a4c2218551e602b6dbc6abb6dee25bb.

PR #152, opened 2024-04-24
Geo point generator refactor

Aderfish:
A small refactor to make the methods of this generator static, and to remove mentions of posts, since we are just generating positions.

### DIFF
--- a/test/mocks/data/geopoint.dart
+++ b/test/mocks/data/geopoint.dart
@@ -90,7 +90,7 @@ class GeoPointGenerator {
       );
     });
 
-    // Generate position positions that are not in the range.
+    // Generate positions that are not in the range.
     final positionsNotInRange = List.generate(outRange, (i) {
       double dDirection = minDiagonalOutRange + distanceBetweenpositions * i;
       return GeoPoint(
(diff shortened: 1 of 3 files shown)


test/mocks/data/geopoint.dart
```diff
@@ -68,28 +76,29 @@ class GeoPointGenerator {
...
       return GeoPoint(
         userLatitude + dDirection,
         userLongitude + dDirection,
       );
     });
 
-    // Generate post positions that are not in the range.
-    final postsNotInRange = List.generate(outRange, (i) {
-      double dDirection = minDiagonalOutRange + distanceBetweenPosts * i;
+    // Generate position positions that are not in the range.
```
TheTexanCodeur, line 93: I think that there is a repetition here. 
```suggestion
    // Generate positions that are not in the range.
```

TheTexanCodeur (approved):
LGTM! 

I suggested a minor change but overall I think that this is a good refractor.

yoannLafore (approved):
Thanks for the refactoring !
The PR looks good to me.

Good job :+1:

PR #154, opened 2024-04-24
Create challenge model

Aderfish:
This PR creates the challenge model and a challenge repository service. Addresses #140.

For now a challenge is simply a post id, an `isCompleted` tag and an expiry time. Challenges are refreshed at most daily for now.

Here are the main features:
1. **Database integration** : Challenges are stored in documents with their post ids for easy retrieval. Challenges are stored in two sub collections, `challenges` and `pastChallenges`, of a parent user document. They contain active challenges and past challenges respectively. Past challenges are remembered so that they cannot reappear. These sub collections attach to user documents for now, but it will not be hard to attach them to other objects, like groups, in the future.
2. **Completed challenges** : Current challenges may contain completed challenges that have not expired yet. That way a user can see which of his daily challenges he has completed and how many he should still do.
3. **Expiration** : Expired challenges get moved to past challenges on update. Challenges pointing to deleted posts also get moved to past challenges on update. Updates happen on every get. Challenges pointing to deleted posts are never cleansed from past challenges. I do not think this is a problem since the point of that collection is to avoid collisions with new challenges.
4. **Repository methods** : The challenge repository service offers methods to complete a challenge and to get current challenges for a user. Getting challenges first deletes invalid current challenges, and then creates new challenges. For now, new challenges are drawn from posts that are between a minimum distance of 500m and a maximum distance of 3km, in no particular order. There can be at most 3 active challenges at a time.

Here are the initial acceptance criteria as present in the issue:
- Dataclasses that represent the data from the database (maybe add a subcollection in the users on Firestore?)
- Repository that provides some functions about the challenges
- Function `refreshIfTooOld` that refreshes the challenges if they were refreshed more than a day earlier (maybe refresh every day at midnight). It should take posts that are more than a distance, but less than another one, while making sure to never stay empty. Be very modular on the functions that chooses the posts, it should for instance be able to only take "hot" posts (based on time and upvotes), and maybe try not to use twice the same post as a challenge). 
- Function that allows to fetch the position and id of the posts (maybe a class `ForeignPost`?) representing a challenge

I am very happy to hear your comments, I am sure there are things that can be improved as this is my first time working on the model part.

Note: the post repository has been `getNearPosts` method expanded to get nearby posts between a minimum radius and a maximum radius around a position.


lib/services/database/challenge_repository_service.dart
```diff
@@ -0,0 +1,175 @@
...
+        postId: post,
+        data: ChallengeData(
+          isCompleted: false,
+          expiresOn: Timestamp.fromDate(expiresOn),
+        ),
+      );
+
+      await challengesCollectionRef
+          .doc(post.value)
+          .set(newChallenge.data.toDbData());
```
yoannLafore, line 124: I think we might want to do a little private function `_addChallenge(ChallengeFirestore challenge)` for this to improve code readibility

lib/services/database/challenge_repository_service.dart
```diff
@@ -0,0 +1,175 @@
...
+      activeChallenges.add(newChallenge);
+      activePostIds.add(post);
+    }
+
+    return activeChallenges;
+  }
+
+  /// moves the challenge with id [pid] from the active challenges to the past challenges
+  Future<void> _deleteChallenge(
```
yoannLafore, line 133: Good use of `batch()` writes here :+1: 
I wonder if calling the function something like `_moveToPastChallenge` would be more explicit.
What do you think ?

lib/services/database/challenge_repository_service.dart
```diff
@@ -0,0 +1,175 @@
...
+
+    possiblePosts = possiblePosts.where(
+      (post) =>
+          geoFirePosition.distanceBetweenInKm(
+            geopoint: post.location.geoPoint,
+          ) >
+          minChallengeRadius,
+    );
+
+    return possiblePosts.map((e) => e.id);
```
yoannLafore, line 164: Just to be a little more explicit:
```suggestion
    return possiblePosts.map((post) => post.id);
```

yoannLafore (commented):
Good job !
This is a lot of pretty complex logic that you have implemented.
This will really be helpful to fully implement the challenge feature !

I have made a few suggestions below.

Also, from what I understand, we are considering expired challenges to be past challenges even though they haven't been completed.
I was wondering if it wouldn't be better to consider as past challenges only the ones that have been completed.
Like that, if the user doesn't complete the challenges for a long period of time, all the posts near him wont be put into the `pastChallenges` collection and the user will still be able to complete those.
What do you think of this ?

Overall, you did a great job !

lib/services/database/challenge_repository_service.dart
```diff
@@ -0,0 +1,175 @@
...
+    final sum = now.add(maxChallengeDuration);
+    final expiresOn = DateTime(
+      sum.year,
+      sum.month,
+      sum.day,
+    ); // truncates to the day
+
+    final possiblePosts = await inRangeUnsortedPosts(pos);
+    final postIt = possiblePosts.iterator;
+    final activePostIds = activeChallenges.map((e) => e.postId).toSet();
```
TheTexanCodeur, line 106: Like @yoannLafore said, we could add a little bit of clarity also here.
```suggestion
    final activePostIds = activeChallenges.map((challenge) => challenge.postId).toSet();
```

TheTexanCodeur (commented):
Good job!
 
I generally agree with the comments made by @yoannLafore and I think that this PR will be ready after you address those.

lib/services/database/challenge_repository_service.dart
```diff
@@ -0,0 +1,175 @@
...
+      maxChallengeRadius,
+    );
+
+    possiblePosts = possiblePosts.where(
+      (post) =>
+          geoFirePosition.distanceBetweenInKm(
+            geopoint: post.location.geoPoint,
+          ) >
+          minChallengeRadius,
+    );
```
JoachimFavre, line 162: I personally think that it would be better to add a `minRadius` optional parameter (and its documentation) to the `_postRepositoryService.getNearPosts`. I'm feeling it would be more concise, more readable, and a better separation of concerns.

test/models/database/challenge/challenge_test.dart
```diff
@@ -0,0 +1,18 @@
+import "package:cloud_firestore/cloud_firestore.dart";
+import "package:flutter_test/flutter_test.dart";
+import "package:proxima/models/database/challenge/challenge_data.dart";
+
+void main() {
+  test("Hash and == works", () {
+    final now = Timestamp.now();
+    ChallengeData a = ChallengeData(isCompleted: false, expiresOn: now);
```
JoachimFavre, line 8: I have two questions on this file:
- Why is named `challenge_test` and not `challenge_data_test`?
- Why is there no test for `ChallengeFirestore` `==` and `hashCode` too? (Note that this question is not applicable if you decided to merge `ChallengeFirestore` and `ChallengeData` as I proposed earlier)

test/services/database/challenge_repository_test.dart
```diff
@@ -0,0 +1,206 @@
...
+  PostRepositoryService postRepository,
+  GeoPoint pos,
+  int n,
+) async {
+  final fakePosts = PostDataGenerator.generatePostData(n);
+  for (final post in fakePosts) {
+    await postRepository.addPost(post, pos);
+  }
+  return fakePosts;
+}
```
JoachimFavre, line 27: I'm feeling this could have its place in other tests too. I think it could make sense to move it to a util file for tests. For instance, why not as a method to `PostDataGenerator`.

test/services/database/challenge_repository_test.dart
```diff
@@ -0,0 +1,206 @@
...
+
+  const userPos = userPosition1;
+  final inChallengeRange = GeoPointGenerator.createOnEdgeInsidePosition(
+    userPos,
+    ChallengeRepositoryService.maxChallengeRadius,
+  );
+
+  final uid = testingUserFirestoreId;
+
+  group("ChallengeRepositoryService", () {
```
JoachimFavre, line 51: I'm not fund of this group name. I don't think we used PascalCase for test naming elsewhere. Moreover, putting all the tests of the file in a group seems to diminish the concept of group in my opinion. I think it is worth trying to split it up, though I'm not sure I see a good way to do it, but maybe the following could make sense:
- Challenge repository getter
- Challenge repository challenge update
- Challenge repository finish and delete

test/services/database/challenge_repository_test.dart
```diff
@@ -0,0 +1,206 @@
...
+
+        final actualPost = await postRepository.getPost(challenge.postId);
+        expect(fakePosts.contains(actualPost.data), true);
+      }
+    });
+
+    test("Complete a challenge", () async {
+      await addPosts(postRepository, inChallengeRange, 1);
+      final challenges = await challengeRepository.getChallenges(uid, userPos);
+
```
JoachimFavre, line 74: I think it is worth verifying also here that the challenge list has length 1, since you also do it later:

```suggestion
  expect(challenges.length, 1);

```

test/services/database/challenge_repository_test.dart
```diff
@@ -0,0 +1,206 @@
...
+        postRepository,
+        inChallengeRange,
+        ChallengeRepositoryService.maxActiveChallenges,
+      );
+
+      final challenges = await challengeRepository.getChallenges(
+        uid,
+        userPos,
+      );
+      final postIds = challenges.map((e) => e.postId).toSet();
```
JoachimFavre, line 110: To complete the triforce, my turn do make this suggestion! :D
```suggestion
      final postIds = challenges.map((challenge) => challenge.postId).toSet();
```

test/services/database/challenge_repository_test.dart
```diff
@@ -0,0 +1,206 @@
...
+
+      for (int i = 0; i < 10; i++) {
+        final challenges =
+            await challengeRepository.getChallenges(uid, userPos);
+
+        expect(challenges.length, 1);
+      }
+    });
+
+    test("Challenge posts are unique", () async {
```
JoachimFavre, line 99: You don't test whether a test from `oldChallenges` can re-appear or not. I think it would also make sense (in this test, or in another).

test/services/database/challenge_repository_test.dart
```diff
@@ -0,0 +1,206 @@
...
+        userPos,
+      );
+      final postIds = challenges.map((e) => e.postId).toSet();
+      expect(postIds.length, challenges.length);
+    });
+
+    test("Deleted post disappears from challenges", () async {
+      final fakePosts = PostDataGenerator.generatePostData(1);
+      final postId =
+          await postRepository.addPost(fakePosts.first, inChallengeRange);
```
JoachimFavre, line 117: why not using `addPosts` with `n=1`? Moreover, I think using `n=2` would make more sense: you verify that `deletePost` deletes one post, but not all of them (and maybe also check it deletes the correct one).

test/services/database/challenge_repository_test.dart
```diff
@@ -0,0 +1,206 @@
...
+          ChallengeRepositoryService.maxActiveChallenges;
+
+      await addPosts(
+        postRepository,
+        inChallengeRange,
+        totalChallenges,
+      );
+      final now = DateTime.now();
+      final past =
+          now.subtract(ChallengeRepositoryService.maxChallengeDuration);
```
JoachimFavre, line 142: I feel this might be a bit dangerous to be exactly at `now - maxChallengeDuration`. I therefore propose instead:
```suggestion
      final past = now.subtract(
        ChallengeRepositoryService.maxChallengeDuration +
            const Duration(seconds: 1),
      );
```

test/services/database/challenge_repository_test.dart
```diff
@@ -0,0 +1,206 @@
...
+
+      final postDocs =
+          await firestore.collection(PostFirestore.collectionName).get();
+      final postIds =
+          postDocs.docs.map((e) => PostIdFirestore(value: e.id)).toList();
```
JoachimFavre, line 147:
```suggestion
          postDocs.docs.map((post) => PostIdFirestore(value: post.id)).toList();
```

test/services/database/challenge_repository_test.dart
```diff
@@ -0,0 +1,206 @@
...
+            .doc(postIds[i].value)
+            .set(challenge.data.toDbData());
+      }
+
+      final updatedChallenges = await challengeRepository.getChallenges(
+        uid,
+        userPos,
+      );
+
+      final updatedPostIds = updatedChallenges.map((e) => e.postId).toSet();
```
JoachimFavre, line 171:
```suggestion
      final updatedPostIds = updatedChallenges.map((challenge) => challenge.postId).toSet();
```

JoachimFavre (commented):
This was a very good PR. I have made some few comments. Overall it is very good, good job for your first contact with the model! :)

lib/services/database/challenge_repository_service.dart
Aderfish: Yeah that's a good idea, I'll keep this in this pr since its a small change, and update the pr description.
I did that in fd198fc4340bbb7db339f62b8d1a5307963d0e1b and 4f7cd91af353f1fa20a73854ab213e168a134f39

lib/services/database/challenge_repository_service.dart
```diff
@@ -0,0 +1,175 @@
...
+    DocumentReference parentRef,
+    GeoPoint pos,
+  ) async {
+    final challengesCollectionRef =
+        parentRef.collection(ChallengeFirestore.subCollectionName);
+    final pastChallengesCollectionRef = parentRef
+        .collection(ChallengeFirestore.pastChallengesSubCollectionName);
+
+    final challengesSnap = await challengesCollectionRef.get();
+    final pastChallengesSnap = await pastChallengesCollectionRef.get();
```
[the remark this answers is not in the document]
Aderfish: Good idea, I added this in a4082d535d89968c9869cf03b0ae22224c6cf8af

lib/services/database/challenge_repository_service.dart
Aderfish: Good idea, I do this in d255627a60235693516231d4f01341875222dcb2

lib/models/database/challenge/challenge_firestore.dart
```diff
@@ -0,0 +1,49 @@
+import "package:cloud_firestore/cloud_firestore.dart";
+import "package:flutter/foundation.dart";
+import "package:proxima/models/database/challenge/challenge_data.dart";
+import "package:proxima/models/database/post/post_id_firestore.dart";
+
+/// This class represents a challenge in the Firestore database
+
+@immutable
+class ChallengeFirestore {
```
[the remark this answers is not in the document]
Aderfish: Just following what I saw for the users and posts, but I agree that it could be merged here.

test/services/database/challenge_repository_test.dart
Aderfish: The actual duration subtracted is irrelevant here. What is important is that past < now. We could probably get away with using now since getting the challenges is done a bit later. Do you think I should use another duration like 1 second to prevent confusion ?

lib/models/database/challenge/challenge_firestore.dart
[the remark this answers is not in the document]
yoannLafore: I personnaly think that it is better to keep the separation for consistency with the rest of the codebase.

lib/services/database/challenge_repository_service.dart
```diff
@@ -0,0 +1,175 @@
...
+    final userDocRef =
+        _firestore.collection(UserFirestore.collectionName).doc(uid.value);
+
+    return _getChallenges(userDocRef, pos);
+  }
+
+  Future<List<ChallengeFirestore>> _getChallenges(
+    DocumentReference parentRef,
+    GeoPoint pos,
+  ) async {
```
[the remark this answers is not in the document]
Aderfish: Good idea, I did this with 97b7664cc57fbfbd1c65ffdbbc046e37b121f1ec and eb66e7688111e0be96166ece9af3f3d18371f3b8

test/services/database/challenge_repository_test.dart
JoachimFavre: Oh, yes ok indeed. Yes, I think using one second would make more sense than using the unrelated maxChallengeDuration! :)

lib/models/database/challenge/challenge_firestore.dart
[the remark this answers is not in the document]
JoachimFavre: The challenge model does not have such a separation

test/services/database/challenge_repository_test.dart
Aderfish: Good idea, I added it and found a bug! 1c6ff44a554f25fd5a30c50d368ea7d9a3226316

test/services/database/challenge_repository_test.dart
Aderfish: You're right, I was being a bit lazy because I first wrote addPosts to only return the post data and not the ids. I added another method to return a list of PostFirestore after adding n. It should be easier to write more tests now.
73a45e83d9c53307dbb6ba9ce6f79faf301ebf90

I changed the test to use two posts as well. 3dcc90c25d88e9e705ee15990b44d291c1490966

Aderfish:
Thank you for the reviews and all the constructive comments. I incorporated them all, tell me if you still see something wrong.

yoannLafore (approved):
Amazing job !
Thanks for bringing the suggestions, I think that the code is now very clean and the PR looks good to me :+1: 

Not extremely important, but I wonder if we could test the `hashcode` and `==` of `ChallengeFirestore` and `ChallengeData` because we do it for other such classes in our tests. We can also move this to another small PR otherwise.

Overall, you did an incredible work implementing the challenge logic which wasn't trivial at all.
So, thank you for your great contribution :tada:

lib/models/database/challenge/challenge_firestore.dart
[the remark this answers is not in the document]
yoannLafore: I don't think that it is striclty necessary to have the separation here.
But I believe that it's nice to keep it for separation of concerns.

lib/services/database/challenge_repository_service.dart
```diff
@@ -0,0 +1,222 @@
...
+  Future<void> completeChallenge(
+    UserIdFirestore uid,
+    PostIdFirestore pid,
+  ) async {
+    final userDocRef =
+        _firestore.collection(UserFirestore.collectionName).doc(uid.value);
+    await _completeChallenge(userDocRef, pid);
+  }
+
+  Future<void> _completeChallenge(
```
JoachimFavre, line 55: By humble curiosity, what is the purpose of this private method? Why isn't it inlined?

lib/services/database/challenge_repository_service.dart
```diff
@@ -0,0 +1,222 @@
...
+    UserIdFirestore uid,
+    GeoPoint pos,
+  ) async {
+    final userDocRef =
+        _firestore.collection(UserFirestore.collectionName).doc(uid.value);
+
+    return _getChallenges(userDocRef, pos);
+  }
+
+  Future<List<ChallengeFirestore>> _getChallenges(
```
JoachimFavre, line 82: Same question here.

lib/services/database/challenge_repository_service.dart
```diff
@@ -0,0 +1,222 @@
...
+    final sum = now.add(maxChallengeDuration);
+    final expiresOn = DateTime(
+      sum.year,
+      sum.month,
+      sum.day,
+    ); // truncates to the day
+
+    final Iterable<PostIdFirestore> possiblePosts =
+        (await _inRangeUnsortedPosts(pos))
+            .where((post) => !justExpired.contains(post));
```
JoachimFavre, line 143: Is this line really needed, all `justExpired` challenges are stored in the `pastChallenges`, aren't they? I'm not sure I understand the purpose of `justExpired` more generally

test/mocks/data/challenge_data.dart
```diff
@@ -0,0 +1,21 @@
+import "package:cloud_firestore/cloud_firestore.dart";
+import "package:proxima/models/database/challenge/challenge_data.dart";
+
+class ChallengeGenerator {
+  ChallengeGenerator._();
+
+  /// Generates a challenge
+  /// by default, the challenge is not completed and expires now
+  /// [isCompleted] whether the challenge is completed
+  /// [expirationDelay] the delay until the challenge expires. The challenge will expire at DateTime.now() + expirationDelay
```
JoachimFavre, line 10: This does not meet Dart's "DO use prose to explain parameters, return values, and exceptions" [documentation convention](https://dart.dev/effective-dart/documentation#do-use-prose-to-explain-parameters-return-values-and-exceptions). Here is a potential fix:
```suggestion
  /// Generates a challenge (using mock data).
  /// It expires in [expirationDelay] (i.e., it expires at `DateTime.now() + expirationDelay`), 
  /// and the generated challenge may be completed or not, depending on [isCompleted].
```

test/mocks/services/mock_post_repository_service.dart
```diff
@@ -53,9 +53,13 @@ class MockPostRepositoryService extends Mock implements PostRepositoryService {
   }
 
   @override
-  Future<List<PostFirestore>> getNearPosts(GeoPoint? point, double? radius) {
+  Future<List<PostFirestore>> getNearPosts(
+    GeoPoint? point,
+    double? maxRadius, [
+    double? minRadius = 0.0,
```
JoachimFavre, line 59: I think this should be tested in `post_repository_test.dart`.

lib/models/database/challenge/challenge_firestore.dart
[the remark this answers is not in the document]
JoachimFavre: I'm personally feeling it adds unnecessary boilerplate and complexity to the code. I guess it will be @Aderfish's choice to change it or not.

test/services/database/challenge_repository_test.dart
```diff
@@ -0,0 +1,252 @@
...
+import "package:proxima/services/database/challenge_repository_service.dart";
+import "package:proxima/services/database/post_repository_service.dart";
+
+import "../../mocks/data/challenge_data.dart";
+import "../../mocks/data/firestore_user.dart";
+import "../../mocks/data/geopoint.dart";
+import "../../mocks/data/post_data.dart";
+
+/// Add [n] posts at position [pos] and return their data and the [PostFirestore] objects
+Future<List<PostFirestore>> addPostsFull(
```
JoachimFavre, line 19: I personally find this function name ambiguous. I'd suggest `addPostsReturnFullPost` (courtesy of Copilot) for instance. This may make sense to rename `addPost` to `addPostsReturnPostData` too.

JoachimFavre (commented):
Great modifications! 
You just forgot to address some of my comments from my original review. Reading what you added, I also added some other minor comments below.
This PR will get merged soon, great job! :)

JoachimFavre (changes requested):
(I think my review makes more sense as "Request Changes", sorry for the inconveniance)

lib/models/database/challenge/challenge_firestore.dart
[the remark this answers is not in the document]
yoannLafore: Effectively it does,  I'm also ok with having a single class, I guess @Aderfish will decide,

lib/services/database/challenge_repository_service.dart
Aderfish: The idea is to make it simple to reuse with group documents later on.

lib/services/database/challenge_repository_service.dart
Aderfish: You're right, it is a relic of a time when I was trying to read both collections before writing to them, because I was trying to use transactions. I removed `justExpired` completely. 57a97123fb4eda47c02905c1753477766ca7cbdc

test/mocks/services/mock_post_repository_service.dart
Aderfish: You're right. Done in 399ed8adda6ef7259ebf18ac451ed88e90355e73

lib/models/database/challenge/challenge_firestore.dart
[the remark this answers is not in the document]
Aderfish: I'll leave it as is.

test/models/database/challenge/challenge_test.dart
Aderfish: I changed the names and added a test for ChallengeFirestore. 4696efafb6b5b742a7e079bbde36ebb16b4f88e8

test/services/database/challenge_repository_test.dart
Aderfish: Done with 1c06dc7e0ba2671444797c58ed83e9a7de88ccc1

test/services/database/challenge_repository_test.dart
Aderfish: I don't really like it in the end. I prefer what we have in the post_repository_test, with full post generation and then setting the database directly. We could maybe move the `setPostFirestore` and `setPostsFirestore` methods defined there to a util file instead, but I'm not sure where.

test/mocks/services/mock_post_repository_service.dart
Aderfish: 399ed8adda6ef7259ebf18ac451ed88e90355e73

TheTexanCodeur (approved):
Great work!

Thank you for taking into account our comments. This PR is ready to be merged in my opinion.

lib/services/database/challenge_repository_service.dart
JoachimFavre: But this will be another repository anyway, won't it? We will have a repository for handling votes on comments, and one for handling votes for posts; why would it be different here?

test/services/database/challenge_repository_test.dart
JoachimFavre: I allow myself to insist, I don't think those functions have their place here. There is a `testutils/` package in which you can put such functions, but I'm feeling this would not be great to find those functions later.
I allow myself to insist, I think they have their place in the `PostDataGenerator`.

lib/services/database/challenge_repository_service.dart
Aderfish: Alright I merged the private generic `_getChallenges` and `_completeChallenge` methods with the public user ones. The helper private methods still think in terms of parent refs but I think that's completely fine and does not decrease readability. c5e5099aa080cf9ec1d1fb0433ad9f24b3a54821

test/services/database/challenge_repository_test.dart
Aderfish: Done in 4793ee6f55dbb6c2057b062d809c160ee89dd7b1 and 3a976865bc1ac79d87849f68e4950d28dea02bd4. I changed the implementation to be inside the post generator and to use setPost. That way the ids simply increase starting from 0, and that will make debugging failing tests easier. So I have also added some methods to the post generator and moved setPost from the post test to the generator file that way we can reuse it too.

test/mocks/data/firestore_post.dart
```diff
@@ -68,4 +77,52 @@ class FirestorePostGenerator {
...
+    return fakePosts;
+  }
+
+  /// Add [n] posts at position [location] and return their data
+  Future<List<PostData>> addPostsReturnDataOnly(
+    FirebaseFirestore firestore,
+    GeoPoint location,
+    int n,
+  ) async {
+    return (await addPosts(firestore, location, n)).map((e) => e.data).toList();
```
JoachimFavre, line 98: I think this can have a slightly improved readability:
```suggestion
    final posts = await addPosts(firestore, location, n);
    return posts.map((post) => post.data).toList();
```

JoachimFavre (approved):
You did a very good job. I just have a final very small suggestion you can find below.

Good job for finishing this PR! This was not trivial but we're finally at the end. Again, you did a great job, and congratulations for making such a good quality code when it is your first contact with the model! :)

PR #155, opened 2024-04-24
Challenge page

### LINKED ISSUE #139: Make the UI for the challenge page
Acceptance criteria:
- Implement the UI from penpot

JoachimFavre:
Fixes #139.

This PR adds the UI for the challenge page:
<p align="center">
  
  
</p>

The acceptance criteria was to implement what was on PenPot, but it appeared its content needed a refactor to match closer what we had in mind. The PenPot has therefore also been updated.

Note that, as part of this PR, I added some `KeyValueListGenerator` class. It will probably not be used elsewhere, but I found its abstraction userful for better understanding my widget. I'm not entirely sure of the way I implemented it, even though I'm pretty happy on how it ended; just keep this in mind while reviewing.

Also, note that I took a strong inspiration from the feed page. I don't think there is major code repetition, but I think it is worth also keeping this in mind while reviewing.


lib/models/ui/challenge_card_data.dart
```diff
@@ -0,0 +1,41 @@
...
+  /// is `null`, the challenge is a group challenge.
+  const ChallengeCardData({
+    required this.title,
+    required this.distance,
+    required this.timeLeft,
+    required this.reward,
+  });
+
+  bool get isFinished => distance == null;
+  bool get isGroupChallenge => timeLeft == null;
```
Aderfish, line 24: I think you should use extra class attributes instead of mixing this state with distance and time left.

We might decide that group challenges also expire, so a challenge being a group challenge would be independent of time left.

For distance we might temporarily lose location service but still want to display challenges that have been generated previously and that may or may not be complete.

Whether a challenge is completed or not, or is a group challenge or not can be obtained from the database.

Aderfish (commented):
I would suggest moving the `views/home_content/challenge` package to `views/pages/challenge`.
To be honest I'm not sure I understand why `home_content` is not in `pages/home` as well so I might be wrong. In any case I'm happy to hear why you chose to put your package there.

Edit: resolved, I now understand better how navigation works thank you @CHOOSEIT
(I should not have linked this with the other comments on the files)

lib/views/home_content/challenge/challenge_card.dart
```diff
@@ -0,0 +1,84 @@
...
+    );
+
+    if (!challenge.isFinished) {
+      listGenerator.addPair("Distance to post", "${challenge.distance} meters");
+    } else {
+      listGenerator.addLine("Challenged finished!");
+    }
+    if (!challenge.isGroupChallenge) {
+      listGenerator.addPair("Time left", "${challenge.timeLeft} hours");
+    }
```
Aderfish, line 31: Adding on to my previous comment, you would rather check if time left is not null here, and maybe show something like unknown if distance is null and the challenge is not finished.

CHOOSEIT:
> I would suggest moving the `views/home_content/challenge` package to `views/pages/challenge`. To be honest I'm not sure I understand why `home_content` is not in `pages/home` as well so I might be wrong. In any case I'm happy to hear why you chose to put your package there.

This differentiate between pages that needs to be pushed on top of the current screen (`views/pages`), from the one that are just are replacing the `body` of the `home` scaffold (`views/home_content`).

Aderfish (changes requested):
Good job in general! I have left a few comments, the main one being making the finished and group state of a challenge independent from position and time left.

lib/views/home_content/challenge/challenge_card.dart
```diff
@@ -0,0 +1,84 @@
...
+    listGenerator.addPair("Reward", "${challenge.reward} Centauri");
+
+    final icon = Opacity(
+      opacity: 0.8, // to make the icon less prominent
```
CHOOSEIT, line 35: The PR is really clean so I try to be a little more picky here :) 

To keep consistency, I think that we can extract the `0.8` into a private static const as you did with `_opacityWhenChallengedFinished` 😄

test/models/ui/challenge_card_data_test.dart
```diff
@@ -0,0 +1,43 @@
...
+        reward: 100,
+      );
+      const challengeCardDataCopy = ChallengeCardData(
+        title: "title",
+        distance: 50,
+        timeLeft: null,
+        reward: 100,
+      );
+
+      expect(challengeCardData, challengeCardDataCopy);
```
CHOOSEIT, line 40: It seems that if we test equality with `expect` this part does not seems to be covered.

(This is also something that Yoan and I encountered in our PR. I think that dart is doing some optimizations and consider both objects as `identical`). The way I solved it was by testing inequalities on my object (73a13a8). I don't think it is important now, but it can make sure that our equalities is working well.

image

test/views/home_content/challenge_list_test.dart
```diff
@@ -0,0 +1,76 @@
...
+        // The parameter findRichText from find.text does not appear to work
+        // so I'm forced to use this uglier method.
+        expect(
+          find.byWidgetPredicate(
+            (widget) =>
+                widget is RichText && widget.text.toPlainText().contains(text),
+          ),
+          findsOneWidget,
+        );
+      }
```
CHOOSEIT, line 53: We could maybe extract this into a "test utils" file as we may want to use it later on

CHOOSEIT (commented):
Besides small comments, good job. Thank you !

test/models/ui/challenge_card_data_test.dart
Aderfish: Replacing const with var or final might also fix this create and two objects in memory.

CHOOSEIT: I just tried and it seems that our dart code analyser does not like it. It prefer to use of `const` for performances...

I think this may make the code not pass on the CI. Which is sad. It was a good idea !

lib/utils/ui/key_value_list_generator.dart
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  /// Adds a new [key]-[value] pair to the text. The value will be bold.
+  KeyValueListGenerator addPair(String key, String value) {
+    _newLine();
+    content.add(TextSpan(text: "$key: ", style: style));
+    content.add(
+      TextSpan(
+        text: value,
+        style: style?.copyWith(fontWeight: FontWeight.w700),
```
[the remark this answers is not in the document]
JoachimFavre: I hadn't realised they were aliases, I should have spent more time in the documentation. Thanks! :)
Fixed in 25749a4d010bf36eecc6c8fa4c82c291c73bd305.

test/views/home_content/challenge_list_test.dart
JoachimFavre: Indeed!
Fixed in 3927fe3a441aa92d9e11c8839d929b15c49a6641.

test/models/ui/challenge_card_data_test.dart
JoachimFavre: After some discussions with @CHOOSEIT, we simply decided to remove the `identical` function, since it will, in fact, almost always return false. This follows [Dart's documentation](https://dart.dev/tools/linter-rules/hash_and_equals).
Fixed in fffdb4d9159282864705a89c1bdcf8590d12e96f.

lib/views/home_content/challenge/challenge_card.dart
JoachimFavre: I'm not sure I agree, even after the rest of the refactor. Those values depend whether this is a `challenge` that is finished or not, and if it is a group one or not. We know those values can't be null by construction, and it makes sense `ChallengeCardData` is tightly linked to the `ChallengeCard` ui in my opinion.
What do you think?

JoachimFavre:
The refactoring was non-trivial, it was hard to find a coherent class structure. This was a discussion done with multiple members of the group, and we decided to go with @yoannLafore's solution, to use a different constructor for each possibilities. I documented our discussion in `challenge_card_data.dart` in case we need it in the future. I mirror it here for simplicity, since I think it is also important for this PR:
```
/// This class was hard to design. We considered multiple possible implementations,
/// and decided to use this one. If we add a parameter requiring to double the number of
/// constructors once more, it should be changed. Here are all the possibilities:
/// 1) Use one constructor per possibility, storing unused parameter as null internally (it is
///    the possibility we chose here).
/// 2) Use a single constructor, asking the developer instanciating it to set the correct
///    parameters to null (a null distance to finish the challenge, for instance).
/// 3) Use a single constructor, but with additional boolean parameters to specify if the
///    class is a group challenge or if it is finished. This requires asserts to check that
///    the parameters are consistent (distance must be null if and only if isFinished is true).
/// 4) Use a single constructor taking a title and a reward, and add methods .notFinished(distance)
///    and .notGroup(timeLeft) to create new instances with the new parameters (in a factory-pattern way).
```

lib/models/ui/challenge_card_data.dart
JoachimFavre: See my comment below, for the refactoring (done in 77dc241bcca498f09c546c57e4dac1524c3bfe9a). 

> For distance we might temporarily lose location service but still want to display challenges that have been generated previously and that may or may not be complete.
For now, we can just keep the previous GPS position in memory, this is very fine and probably the easiest way to solve many problems when loosing the internet connection. However, if we want to change this in the future, this can be easily done with the refactor (we would, in this case, indeed need some `isFinished` boolean parameter).

lib/views/home_content/challenge/challenge_card.dart
Aderfish: Good job making this clearer by adding constructors for each case.

Aderfish (approved):
Good job implementing the suggestions.

CHOOSEIT (approved):
Looks good to me ! 
Great PR, clean job

Thank you ! 🎉

PR #158, opened 2024-04-25
User avatar component

CHOOSEIT:
Fixes: #157

This PR aims to provide a user avatar component that can be used easily.

Acceptance criteria:

- [x] Make an abstraction for the UI of the user avatar
- [x] Replace all the occurences in the UI by this
- [x] This is only a refactor and it must not add any functionality, i.e. no need to make it dynamic (to take the real logged in username initial) in places where it is currently static


lib/utils/ui/user_avatar.dart
```diff
@@ -0,0 +1,41 @@
...
+class UserAvatar extends StatelessWidget {
+  const UserAvatar({
+    super.key,
+    required this.displayName,
+    this.radius,
+    this.onTap,
+  });
+
+  final String displayName;
+  final double? radius;
```
camillelnne, line 15: Do think it is a good idea to have `radius` accept `null` as a parameter? 
When instanciating CircleAvatar, it will then default to 20 pixels (cf [here](https://api.flutter.dev/flutter/material/CircleAvatar/radius.html),) but it might be better to force the developer to set a radius, so that they follow the mockup.

Else we can also adapt the default size, for example:
`return CircleAvatar(
    radius: radius ?? (some condition ? value1: value2),
    ...
  );`

camillelnne (commented):
Thank you for fixing this issue!
Will definitely help in the future.
I just left a comment about the radius default size, else LGTM

lib/utils/ui/user_avatar.dart
CHOOSEIT: You are completely right ! 

Following your comment, another solution would have been to use the default parameter as `this.radius = 20` in the constructor.

But I think that forcing the developer to carefully choose the radius according to the mockup and not just taking the default one is important.

Added in: [ec8f523](https://github.com/ProximaEPFL/proxima/pull/158/commits/ec8f523b7d68502d1d1cacf26a4c2091b0799645)

Thank you 😄

camillelnne (approved):
Thanks for the changes!
LGTM :)

lib/views/pages/home/top_bar/app_top_bar.dart
```diff
@@ -19,24 +20,14 @@ class AppTopBar extends HookConsumerWidget {
...
-              key: profilePictureKey,
-              onTap: () => {
-                Navigator.pushNamed(context, Routes.profile.name),
-              },
-            ),
-          ),
-        ],
-      ),
+    //TODO: add user display name to UserAvatar
+    Widget userAvatar = UserAvatar(
```
TheTexanCodeur, line 24: Shouldn't we make userAvatar final like before here?

TheTexanCodeur (commented):
Good job!

I just made a little comment. Could you please justify the choice you made? Except from that I don't have anything else to say.

lib/views/pages/home/top_bar/app_top_bar.dart
CHOOSEIT: You are completely right ! Good catch !  

Fixed in: [6e71c2a](https://github.com/ProximaEPFL/proxima/pull/158/commits/6e71c2a6c832c9d62e200dbc3f34aedfcefa290a)

Thank you. 😄

camillelnne:
I noticed that this is linked to an issue #157, that mentioned the user initials. So maybe we should split it/reword it, because this PR doesn't cover the initials part, right?

TheTexanCodeur (approved):
Thanks for the changes! 

This PR  LGTM now.

lib/utils/ui/user_avatar.dart
```diff
@@ -0,0 +1,41 @@
...
+  const UserAvatar({
+    super.key,
+    required this.displayName,
+    required this.radius,
+    this.onTap,
+  });
+
+  final String displayName;
+  final double radius;
+  final void Function()? onTap;
```
gruvw, line 16: You can use `VoidCallback?` here.

CHOOSEIT:
> I noticed that the title issue #157 mentions the display user initials. The new class is coded with this in mind, but for now when instantiating, we pass hard coded names. So I think the title of the issue is a bit confusing, because I thought it would mean that the initials would be personalized. However, your PR satisfies the acceptance criteria perfectly! So I propose we change the title of the issue and create another one for the personalization, that we can fix later.

You are completely right ! 

I am going to ask the scrum master to change the name of the issue. 

Thank you for letting me know ! 😄

lib/utils/ui/user_avatar.dart
CHOOSEIT: You are right ! I wasn't aware of this typedef, thank you !

Added in: [06bbad6](https://github.com/ProximaEPFL/proxima/pull/158/commits/06bbad6a54607044acbe9f7d67c31a56409fa3d9)

JoachimFavre:
> I am going to ask the scrum master to change the name of the issue.

I edited the description of #157 and added #183 to the product backlog! :)

CHOOSEIT:
I modified the PR description to add the new acceptance criteria. @camillelnne Told me that the changes were good to merge.

PR #179, opened 2024-04-28
Basic map UI

TheTexanCodeur:
This pull request implements a new screen that displays a map. Here are the main features brought with this PR:

1. **Navigation to the map** : the "map" button of the main bottom navigation widget now leads to a new screen with a map and filter chips. 

2. **Mock positon** : for now, the map is loaded at a mock location. On top of that, a circle (100m radius) is drawn around the user.  

3. **Testing** : additional tests are implemented for this new screen. They verifiy that the widgets are displayed, that errors (like no GPS) are handled correctly and that the navigation from the post feed to the map occurs flawlessly.


test/end2end/app_test.dart
```diff
@@ -158,7 +160,8 @@ Future<void> bottomNavigation(WidgetTester tester) async {
   // Map
   await tester.tap(find.text("Map"));
   await tester.pumpAndSettle();
-  expect(find.text("Proxima"), findsOneWidget);
+  expect(find.byType(MapScreen), findsOneWidget);
+  expect(find.byType(PostMap), findsOneWidget);
```
yoannLafore, line 164: I think that only checking that `MapScreen` is present is enough here because this test only checks at a high level that elements are displayed.
This is only my interpretation though, so what do you think ?

```suggestion
```

test/mocks/overrides/override_map_view_model.dart
yoannLafore, line 1: Good initiative to create a Mock for testing :+1: 
I just think that this implementation is not exactly what we are expecting for a mock.
I believe we should inject as dependencies the `build()` and `refresh()` functions.
I think you can take inspiration from the `MockHomeViewModel` for the implementation.

lib/views/select_option_widgets/map_selection_option_chips.dart
```diff
@@ -0,0 +1,28 @@
+import "package:flutter/material.dart";
+import "package:proxima/views/select_option_widgets/map_selection_options.dart";
+
+class MapSelectionOptionChips extends StatelessWidget {
+  const MapSelectionOptionChips({
+    super.key,
+  });
+  static const selectOptionsKey = Key("Select options");
```
yoannLafore, line 8: Just camelCase :+1: 
```suggestion
  static const selectOptionsKey = Key("selectOptions");
```

test/mocks/providers/provider_map_page.dart
```diff
@@ -0,0 +1,46 @@
...
+
+ProviderScope newMapPageNoGPS(
+  MockGeoLocationService geoLocationService,
+  MockPostRepositoryService postRepositoryService,
+) {
+  return ProviderScope(
+    overrides: [
+      geoLocationServiceProvider.overrideWithValue(geoLocationService),
+      postRepositoryProvider.overrideWithValue(postRepositoryService),
+      mapProvider.overrideWith(() => MockMapViewModel()),
```
yoannLafore, line 38: If you go with the modification of the `MockMapViewModel`, you might need to modify this to have the wanted behavior.

test/views/pages/map/navigation_map_test.dart
```diff
@@ -0,0 +1,82 @@
...
+        ),
+      );
+
+      GeoPoint testPoint = userPosition0;
+      when(geoLocationService.getCurrentPosition()).thenAnswer(
+        (_) => Future.value(testPoint),
+      );
+
+      await tester.pumpAndSettle();
+    });
```
yoannLafore, line 55: I think you forgot to add an `expect` to test the behavior here.

test/mocks/providers/provider_map_page.dart
yoannLafore, line 1: I don't really understand why we need to have a `geoLocationService` here.
Because the current implementation of the `MapViewModel` is only here as a placeholder and only exposes the mock location `const LatLng(46.519653, 6.632273)`.
Thus I'm not sure we need the `geoLocationService` in this PR.

test/views/pages/map/navigation_map_test.dart
yoannLafore, line 1: Similarly, I don't think that mocking the `geoLocationService` is necessary for similar reasons to my comment above.
Actually, I don't think that the tests need the `geoLocationService` or the `postRepositoryProvider` given that the view model has only be implemented as a placeholder for the real implementation later.

test/views/pages/map/navigation_map_test.dart
yoannLafore, line 55: I just think you forgot to add an `expect` to test the behavior.

test/views/pages/map/navigation_map_test.dart
yoannLafore, line 1: The current implementation of the `MapViewModel` is only here as a placeholder for when the real implementation will be done later.
Thus I don't think we need to use the `geoLocationService` or `postRepositoryProvider` in the tests.
(This also applies to `static_map_test.dart`)

yoannLafore (commented):
The UI implementation of the map screen looks good to me in the overall :+1: 
(I have just made mostly style suggestions on that).

I am just not entirely sure about implementing a placeholder viewmodel `MapViewModel` in this PR.
I personnaly believe that it could have been easier to simply not implement any viewmodel and only use mocked data in the UI.
But this is only my point of view and I guess that we can discuss that further.

Regarding the tests, I am not sure why they are using `geoLocationService` and `postRepositoryProvider` but I have made a comment on that below.

Overall, I believe you did a good job at implementing the map screen.
(I just think that the testing side might need some rework as I detail in my suggestions).

lib/views/select_option_widgets/map_selection_option_chips.dart
```diff
@@ -0,0 +1,28 @@
...
+        label: Text(selectOption.name),
+      );
+    }).toList();
+
+    return Wrap(
+      key: selectOptionsKey,
+      alignment: WrapAlignment.center,
+      spacing: 4,
+      children: selectOptions,
+    );
```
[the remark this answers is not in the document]
TheTexanCodeur: You're right this could also be an option!  We discussed about this with the team and came to the conclusion that a non scrollable presentation would be easier to manage for the user (less scrollable elements). Thanks for the suggestion anyway!

test/end2end/app_test.dart
TheTexanCodeur: I understand your point of view, but I think that there's never enough testing. In my opinion, it is also important to check that map itself was loaded correctly. Even if we're at high level, I think that assuring that the map screen indeed displays a map is crucial.

yoannLafore: Ok I understand, then we can let it like that.

lib/viewmodels/map_view_model.dart
```diff
@@ -0,0 +1,33 @@
...
+      currentLocation: const LatLng(46.519653, 6.632273),
+    );
+  }
+
+  /// Refresh the mapInfo.
+  /// This will put the state of the viewmodel to loading
+  /// and update the state accordingly
+  Future<void> refresh() async {
+    state = const AsyncValue.loading();
+    state = await AsyncValue.guard(() => build());
```
yoannLafore, line 27: While checking the coverage, I noticed that this was not covered by the tests.
This makes perfect sense given that the view model is only a place holder for the future real implementation.
Thus, I was wondering if we shouldn't keep the refresh method empty for this PR and then it will be implemented during the PR for the map view model.
I wonder what are your thoughts on this ?

yoannLafore (commented):
The PR looks good to me in the overall.
I have just made a few aditionnal suggestions below.

lib/viewmodels/map_view_model.dart
TheTexanCodeur: Even if it's not tested yet, I think that the logic of the button should already by implemented, as one of the acceptance criteria implies that the user can retry to load the screen when an error occurs. On top of that, those specific lines will be covered in the next PR's, with the implementation of the real user location arriving soon.

lib/viewmodels/map_view_model.dart
```diff
@@ -0,0 +1,44 @@
...
+  @override
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
+
+    return other is MapInfo && other.currentLocation == currentLocation;
+  }
+
+  @override
+  int get hashCode => currentLocation.hashCode;
+}
```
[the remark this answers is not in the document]
Aderfish: It would make sense for this class to contain the currently selected map option as well, even if they are only implemented in the future.

Aderfish (changes requested):
Good job getting started on the map page! I have left some comments, most of them are small. For the two bigger refactor suggestions that touch other parts of the code, it might make sense to open other issues and address them in other prs.

lib/views/home_content/map/map_screen.dart
```diff
@@ -0,0 +1,58 @@
...
+    final fallback = Center(
+      child: Column(
+        mainAxisAlignment: MainAxisAlignment.center,
+        children: [
+          const Text("An error occurred"),
+          const SizedBox(height: 10),
+          refreshButton,
+        ],
+      ),
+    );
```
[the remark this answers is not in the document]
TheTexanCodeur: You're right! I added this small task to the product backlog.

lib/viewmodels/map_view_model.dart
yoannLafore: This seems good to me then :+1:

lib/views/select_option_widgets/map_selection_options.dart
```diff
@@ -0,0 +1,12 @@
...
+enum MapSelectionOptions {
+  heatMap("Heat map"),
+  nearby("Nearby"),
+  myPosts("My posts"),
+  challenges("Challenges");
+
+  final String name;
+
+  const MapSelectionOptions(this.name);
+}
```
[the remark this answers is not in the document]
TheTexanCodeur: Done in cbc0242

lib/views/home_content/map/post_map.dart
```diff
@@ -0,0 +1,45 @@
...
+        myLocationButtonEnabled: false,
+        myLocationEnabled: false,
+        zoomGesturesEnabled: true,
+        zoomControlsEnabled: true,
+        scrollGesturesEnabled: true,
+        rotateGesturesEnabled: false,
+        tiltGesturesEnabled: false,
+        initialCameraPosition: CameraPosition(
+          target: mapInfo.currentLocation,
+          zoom: 17.5,
```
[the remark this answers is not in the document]
Aderfish: I'm sorry I meant to only highlight the zoom level. 17.5 could be named initialZoomLevel, so that we can easily modify it later from the top of the file.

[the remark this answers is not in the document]
TheTexanCodeur: Fixed with 726ef27

lib/views/select_option_widgets/map_selection_option_chips.dart
```diff
@@ -0,0 +1,28 @@
...
+  });
+  static const selectOptionsKey = Key("selectOptions");
+
+  @override
+  Widget build(BuildContext context) {
+    final selectOptions = MapSelectionOptions.values.map((selectOption) {
+      return ChoiceChip(
+        //TODO: Handle the selection of the options
+        key: Key(selectOption.name),
+        selected: selectOption == MapSelectionOptions.values[1],
```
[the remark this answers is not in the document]
Aderfish: Now that you have added the selection option to the view state, you should use it here instead of using a constant.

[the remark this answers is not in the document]
TheTexanCodeur: You're right. Fixed with b17eb0b

lib/models/ui/map_info.dart
```diff
@@ -0,0 +1,23 @@
...
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
+
+    return other is MapInfo &&
+        other.currentLocation == currentLocation &&
+        other.selectOption == selectOption;
+  }
+
+  @override
+  int get hashCode => currentLocation.hashCode;
```
yoannLafore, line 22: I believe the `hashcode` should take also take the `selectedOption`
```suggestion
  int get hashCode => Object.hash(
        currentLocation,
        selectOption,
      );
```

test/views/pages/map/static_map_test.dart
```diff
@@ -41,4 +44,42 @@ void main() {
...
+        selectOption: MapSelectionOptions.nearby,
+      );
+      const mapInfo3 = MapInfo(
+        currentLocation: LatLng(0, 0),
+        selectOption: MapSelectionOptions.nearby,
+      );
+
+      expect(mapInfo1.hashCode, isNot(mapInfo2.hashCode));
+      expect(mapInfo1.hashCode, mapInfo3.hashCode);
+    });
```
yoannLafore, line 65: I am not sure that this is exactly how we want to test `hashcode`.
Because it is not forbidden for two different classes to have the same hash (even if it is very unlikely).
I believe we could follow an implementation similar to the tests for `PostOverview` where we check that the hashcode function computes the hash of its parameters. (those are located under `test/models/ui/post_overview_test.dart`)
Also, I believe those tests (`hashcode` and `==`) should be placed under
`test/models/ui/map_info_test.dart`.

yoannLafore (commented):
Thanks for the modifications !
The PR looks good to me in the overall.
I have just two little suggestions below and then I think that it should be ready to merge for me :+1:

test/views/pages/map/static_map_test.dart
TheTexanCodeur: Fixed with deda72b!

yoannLafore (approved):
The PR looks good to me :+1: 
Thank you for your good work :tada:

Aderfish (approved):
This PR looks good to me now. Good work!

Issue #176, opened 2024-04-26
Create model and repository for comments


JoachimFavre:
Acceptance criteria:
- Make model to comment posts. Only one level of depth.
- Make the repository to expose the comments of a post.
- The repository must expose methods to get/add/delete comments on a posts
- When the number of comments on a post changes, the `commentCount` of the post is updated accordingly

PR #180, opened 2024-04-28
Comment model repository

yoannLafore:
Hello,
This PR aims to provide a a way of exposing the comments of a post (issue #176).
To do so, this PR does the following:

**Provide a representation of a comment:** 
This is done through the classes `CommentFirestore`, `CommentIdFirestore` and `CommentData`.

**Add a comment count on post:**
The `PostData` class now has a `commentCount` field that represent the number of comments under the post.
I also have done such that when the `commentCount` field doesn't exist on the database, the number of comments is interpreted as 0.
This is done through `commentCount: data[commentCountField] ?? 0,` in `PostData.fromDbData`.
I don't know if it is the best solution but given that the database already contains a substantial amount of posts I chose to go with this solution to ensure compatibilty. I'd like to get your thoughts on this.

**Provide a repository to access, add, delete comments on a post:**
The repository is implemented under `CommentReposiroryService`. It provides the following mehods:
- `getComments(parentPostId)`: gets all the comments that are under a post
- `addComment(parentPostId, commentData)`: atomically adds a comment on the post and update the `commentCount` field.
- `deleteComment(parentPostId, commentId)`: atomically deletes a comment on the post and update the `commentCount` field. If the comment doesn't exist, nothing is done (this is to keep consistency of `commentCount`).

**Add tests for the model and repository:**
This PR adds tests for both the model classes and the comment repository.

**Add mock comment repository service:**
This PR adds the class `MockCommentRepositoryService` to be mocked with Mockito for when tests will need to have a mock implementation of the comment repository.

#### Acceptance criteria
- Make model to comment posts. Only one level of depth.
- Make the repository to expose the comments of a post.
- The repository must expose methods to get/add/delete comments on a posts
- When the number of comments on a post changes, the `commentCount` of the post is updated accordingly


gruvw:
Good job of not using inheritance in your new database model types. It was definitely the right choice :+1:

lib/models/database/post/post_data.dart
```diff
@@ -19,12 +19,16 @@ class PostData {
   final int voteScore;
   static const String voteScoreField = "voteScore";
 
+  final int commentCount;
```
gruvw, line 22: How do we handle this change when we will add a comment to a post that does not have this field in the database ?

Is the call to `set` that automatically adds the field on the document on the database ?

lib/services/database/comment_repository_service.dart
```diff
@@ -0,0 +1,124 @@
...
+import "package:proxima/models/database/post/post_firestore.dart";
+import "package:proxima/models/database/post/post_id_firestore.dart";
+import "package:proxima/services/database/firestore_service.dart";
+
+/// This class is a service that allows to interact with the comments
+/// of the posts in the firestore database.
+class CommentRepositoryService {
+  final FirebaseFirestore _firestore;
+
+  CommentRepositoryService({required FirebaseFirestore firestore})
```
gruvw, line 16: To stay consistent with other Services formatting:

```suggestion
  CommentRepositoryService({
    required FirebaseFirestore firestore,
  }) : _firestore = firestore;
```

lib/models/database/comment/comment_data.dart
```diff
@@ -0,0 +1,71 @@
+import "package:cloud_firestore/cloud_firestore.dart";
+import "package:flutter/foundation.dart";
+import "package:proxima/models/database/user/user_id_firestore.dart";
+
+@immutable
+class CommentData {
```
gruvw, line 6: I think we could add a single line of documentation for each of the data classes. Just to inform the user of the class and overall usage of the class in code editors.

lib/models/database/comment/comment_id_firestore.dart
```diff
@@ -0,0 +1,17 @@
+// The comment id is strong typed to avoid misuse
```
gruvw, line 1: This comment should not be at the top of the file. It should be a documentation comment on the class, see:
https://github.com/ProximaEPFL/proxima/blob/3a5188de1129ab86c0f4dad909730778dcf75c55/lib/models/database/post/post_id_firestore.dart#L1-L5

lib/services/database/comment_repository_service.dart
```diff
@@ -0,0 +1,124 @@
...
+    PostIdFirestore parentPostId,
+    CommentIdFirestore commentId,
+  ) async {
+    await _firestore.runTransaction((transaction) async {
+      await _deleteComment(parentPostId, commentId, transaction);
+    });
+  }
+
+  // Concrete implementation of the deletion of a comment
+  Future<void> _deleteComment(
```
gruvw, line 96: I am not sure why we have a private method here, as it is only used once in `deleteComment`.
I assume it is to avoid a lengthy async transaction body, but in my opinion the locality of behavior principle should be applied here.

lib/services/database/comment_repository_service.dart
```diff
@@ -0,0 +1,124 @@
...
+
+  /// This method will delete the comment with id [commentId] from the
+  /// post with id [parentPostId].
+  /// It will also update the number of comments of the post.
+  /// This is done atomically.
+  /// If the comment does not exist, the method will do nothing.
+  Future<void> deleteComment(
+    PostIdFirestore parentPostId,
+    CommentIdFirestore commentId,
+  ) async {
```
gruvw, line 89: Here we can remote the `async` clause of the function and simply return the same `Future` that `_firestore.runTransaction` emits. (like `return _firestore.runTransaction`)

lib/services/database/comment_repository_service.dart
```diff
@@ -0,0 +1,124 @@
...
+    CommentIdFirestore commentId,
+    Transaction transaction,
+  ) async {
+    final commentRef =
+        _commentsSubCollection(parentPostId).doc(commentId.value);
+
+    final commentDoc = await transaction.get(commentRef);
+
+    if (!commentDoc.exists) {
+      return;
```
gruvw, line 107: We probably want to error in that case right ? I don't really like the idea of silently succeeding the operation of removing a comment that does not exist.
Is there a particular reason behind that choice ?

lib/models/database/post/post_data.dart
gruvw: Sorry, instead of `set` I meant `increment`.

test/mocks/data/comment_data.dart
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  static CommentData createRandomCommentData({
+    String? content,
+    UserIdFirestore? ownerId,
+    Timestamp? publicationTime,
+    int? voteScore,
+  }) {
+    return CommentData(
+      content: content ?? "content_${Random().nextInt(100)}",
```
gruvw, line 26: Here we should either pass an instance of `Random()` to the `createRandomCommentData` or have an instantiable `CommentDataGenerator` class that takes an instance of `Random` as an argument (I personally prefer option 2).

Also note that any random testing should be seeded (for reproducibly reasons) so when creating the `Random` instance in testing, please make sure to provide a `const` seed to its constructor parameter.

gruvw: The same goes for `test/mocks/data/firestore_post.dart`, even though I just saw that it was not done before... It must have slipped through the review at the time.

test/services/database/comment_repository_test.dart
```diff
@@ -0,0 +1,249 @@
...
+    });
+
+    group("getting comments", () {
+      test("should return an empty list if there are no comments", () async {
+        final comments = await commentRepository.getComments(postId);
+
+        expect(comments, isEmpty);
+      });
+
+      test(
```
gruvw, line 93: I am not really a fan of that kind of testing where the test basically duplicate the behavior of the supposed component under test, and then we just check that the two codes are actually doing the same thing.
What I prefer is to simply use the component under test and then assert if the resulting behavior is the one we expected (black box testing).
I don't know what @JoachimFavre (other reviewer) thinks about this topic ?

lib/models/database/post/post_data.dart
yoannLafore: That is a really good point indeed !
After some testing, it effectively does add automatically the field if it is missing.

I have added a test for that scenario in bcbe1b262f1c37a83bb573c27febe4ece883bf69.

lib/services/database/comment_repository_service.dart
yoannLafore: Fix in 159a7c3c6ee3839739a78bf7df1736f6d97b150b :+1:

lib/models/database/comment/comment_data.dart
yoannLafore: You're are right, done in 4c0fe6830a9dc7c25e90fb958efa76ed6b5547a5.

lib/models/database/comment/comment_id_firestore.dart
yoannLafore: Completely right, I don't know how I missed it when creating the class.
Fix in 4c0fe6830a9dc7c25e90fb958efa76ed6b5547a5 :+1:

lib/services/database/comment_repository_service.dart
yoannLafore: This was indeed to reduce the body of the run transation and to have a better separation of logic.
I feel like it makes the code easier to read.

Also, I just noticed that using this function I could use the arrow notation to have the `deleteComment` more consice. I do that in 2d01850c7f0221d6121ec633d3c6587df1546db6.

Please tell me what you think about it and if you think it's still better to put it in the transaction body then I'll do it :+1:

lib/services/database/comment_repository_service.dart
yoannLafore: You are right, actually I think we can even use the arrow notiation here.
Thus I use it in 2d01850c7f0221d6121ec633d3c6587df1546db6.

lib/services/database/comment_repository_service.dart
yoannLafore: At first I though it was the best option to simply return.
But thinking again about it I agree with you.
Fix in d0a6592de9d506826ca7439a4206255dfbcabf79.

test/services/database/comment_repository_test.dart
yoannLafore: I am not sure I understand really well.
Because from my point of view I am not comparing that my test code is doing the same as my production code.

I have added some helper functions to facilitate the creation of the conditions for the tests but that's it.
Then I check that I get the expected behavior depending on the condition.

For instance here I check that I indeed get the comments that are on the database:
```dart
      test(
          "should return the comments of a post when there are multiple comments",
          () async {
        final comments = await addComments(5);

        final actualComments = await commentRepository.getComments(postId);

        expect(actualComments, unorderedEquals(comments));
      });
```

Here, `addComments` is only used to put the initial comments on the database. I don't really see how I could do it another way ?

lib/models/database/post/post_data.dart
gruvw: Awesome :)

lib/services/database/comment_repository_service.dart
gruvw: It might be a personal opinion, but I don't really like this syntax :rofl:...

I prefer to keep the function body and have everything in the transaction for the reason explained above. It is not very important if you feel strongly about this.

test/mocks/data/comment_data.dart
gruvw: I agree.

test/services/database/comment_repository_test.dart
gruvw: Yes but that's exactly my point...
Your testing "helper" method replicates the actual implementation and introduces a lot of coupling between the two.
I would use the add comment method from the repository directly and check for correct outside observable behavior instead of comparing it with a "duplicate" method.

lib/services/database/comment_repository_service.dart
yoannLafore: Ok I see :sweat_smile: .
Then I'll move it in the body of `runTransaction` (this actually makes the code more concise).
Done in 7d5766d3590391cfae2ffc7898fa7a480b4cb89c.

test/mocks/data/comment_data.dart
yoannLafore: Nice, I have created the issue #203 for it :+1:

test/services/database/comment_repository_test.dart
yoannLafore: Ohh ok I see.
I think the problem with this approach is that the tests are no longer unitary for the `CommentRepository`.

For instance, the above test would rely on `CommentRepository.addComment` and `CommentRepository.getComments`.

This can be an issue because for instance if `CommentRepository.addComment` would simply store the comments in the memory instead of putting them on firestore, we would have no way of checking that the `getComments` doesn't actually give the comments that are on the database.
That's one of the main reasons for why I chose to use helper functions to ensure that the data are actually coming from the database.

gruvw: It is still unit testing, go with the mantra "Just don't worry about it and test the exposed behaviour". Check out https://stackoverflow.com/questions/5958895/how-can-i-correctly-unit-test-methods-that-depend-on-each-other it's a great thread.

gruvw: Yes exactly :) Thank you very much.

gruvw (approved):
Great work, good feedback and interesting discussions on this PR :rocket:

JoachimFavre (commented):
This is a very good PR, good job. I have some small suggestions hereinbelow, but this should be quick fixes :)

lib/services/database/comment_repository_service.dart
```diff
@@ -0,0 +1,117 @@
...
+
+    return comments;
+  }
+
+  /// This method will add the comment with data [commentData] to the
+  /// post with id [parentPostId].
+  /// It will also update the number of comments of the post.
+  /// This is done atomically.
+  ///
+  /// The method returns the id of the comment that was added.
```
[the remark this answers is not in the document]
yoannLafore: Thanks !

test/mocks/data/firestore_comment.dart
```diff
@@ -0,0 +1,25 @@
...
+import "comment_data.dart";
+
+class CommentFirestoreGenerator {
+  int _commentId = 0;
+  final CommentDataGenerator _commentDataGenerator;
+
+  CommentFirestoreGenerator({int seed = 0})
+      : _commentDataGenerator = CommentDataGenerator(seed: seed);
+
+  CommentFirestore createRandomComment({
```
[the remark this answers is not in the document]
yoannLafore: You're right, fix in ad75041992d5a12e795c7325af7f64807cd4c735.

test/mocks/data/comment_data.dart
```diff
@@ -0,0 +1,38 @@
...
+      return CommentData(
+        content: "content_$i",
+        ownerId: UserIdFirestore(value: "owner_id_$i"),
+        publicationTime: Timestamp.fromMillisecondsSinceEpoch(1000 * i),
+        voteScore: i,
+      );
+    });
+  }
+
+  CommentData createRandomCommentData({
```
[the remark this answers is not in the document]
yoannLafore: Fix in fd564e650fdad48016a179e35d236eecd7e2d4ac.

test/services/database/comment_repository_test.dart
```diff
@@ -0,0 +1,261 @@
...
+      for (var i = 0; i < number; i++) {
+        final commentData = commentDataGenerator.createRandomCommentData();
+        final commentId =
+            await commentRepository.addComment(postId, commentData);
+        comments.add(CommentFirestore(id: commentId, data: commentData));
+      }
+
+      return comments;
+    }
+
```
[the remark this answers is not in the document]
yoannLafore: Good idea !
Done in 86c663eedf1161afcf8b4a466f81f3078800c1f2.

test/services/database/comment_repository_test.dart
```diff
@@ -0,0 +1,261 @@
...
+        firestore: fakeFirestore,
+      );
+
+      // We get the comment repository from the provider container
+      // because it allows to check that the provider constructs
+      // the repository correctly
+      final container = ProviderContainer(
+        overrides: [
+          firestoreProvider.overrideWithValue(fakeFirestore),
+          postRepositoryProvider.overrideWithValue(postRepository),
```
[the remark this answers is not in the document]
yoannLafore: Fix in cf51a0fa6f92189e997ac90152b3de2a4b8e89e7.

test/services/database/comment_repository_test.dart
```diff
@@ -0,0 +1,261 @@
...
+        ],
+      );
+
+      commentRepository = container.read(commentRepositoryProvider);
+
+      final postData = PostDataGenerator().postData;
+      postId = await postRepository.addPost(postData, userPosition0);
```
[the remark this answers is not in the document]
yoannLafore: Fix in 2e5037120625e06ea2ef97a34134dcd4e57ce3b9.

JoachimFavre (approved):
LGTM now, good job ! :)

yoannLafore:
Thank you for great suggestions :tada:

Issue #178, opened 2024-04-28
Add model to vote on comment


yoannLafore:
Acceptance criteria: 
- The service propose a way to upvote/downvote a comment
- When vote is performed, the voteScore of the comment is updated
- The database keeps track of voters on a comment
- The service expose a way to get the vote state of a user on a particular comment

PR #181, opened 2024-04-28
Comment votes

yoannLafore:
Hello,
This PR aims to provide a way of upvoting/downvoting a comment (issue #178).
In order to avoid code duplication, this PR factorises the code from `PostUpvoteRepositoryService` to allow its usage for the comment voting logic.
More concretely, this PR does the following.

**Add an id firestore type:**
This PR adds the abstract class `IdFirestore` that represents a firestore document Id. 
Thus the classes `UserIdFirestore`, `PostIdFirestore` and `CommentIdFirestore` extend this class.
This class is necessary for reasons explained later below.

**Factor the upvote repository logic:**
A new class `UpvoteRepositoryService<ParentIdFirestore extends IdFirestore>` has been created to represent the voting logic that can be applied at multiple places (posts and comments for now).
The logic has been adapted from the initial `PostUpvoteRepositoryService`.

This class aims to provide the voting logic for an abitrary parent document whose id is of type `ParentIdFirestore` (we must insure that `ParentIdFirestore` represents a firestore id thus the introduction of `IdFirestore`). Thus the constructor takes as parameters:
- `firestore`: The firestore instance
- `parentCollection`: The collection in which the parent documents are stored
- `voteScoreField`: The field in the parent document in which the vote score is stored

**Add providers for the post and comment voting logic:**
The provider for the post voting has been adapted to use the new `UpvoteRepositoryService`.
The provider for the comment voting `commentUpvoteRepositoryService` takes as parameter a `postId` and creates the upvote repository for the comments under the post with id `postId`.

**Add tests for the comment voting repository:**
This PR adds some tests for the `commentUpvoteRepositoryService`.

#### Acceptance criteria
- The service propose a way to upvote/downvote a comment
- When vote is performed, the voteScore of the comment is updated
- The database keeps track of voters on a comment
- The service expose a way to get the vote state of a user on a particular comment

**Important note:** Waiting for #180 to be merged for changing base and merging.


test/mocks/services/mock_upvote_repository_service.dart
```diff
@@ -0,0 +1,42 @@
...
+  Future<UpvoteState> getUpvoteState(
+    UserIdFirestore? userId,
+    ParentIdFirestore? parentId, {
+    Transaction? transaction,
+  }) {
+    return super.noSuchMethod(
+      Invocation.method(
+        #getUpvoteState,
+        [userId, parentId],
+        {#transaction: transaction},
```
CHOOSEIT, line 20: It is maybe a bit picky 😄 , but it may be better to keep the consistency of not-named parameter.
```suggestion
        [userId, parentId, transaction],
```

CHOOSEIT (commented):
Looks good to me, expect some small optional suggestions the PR is good to go ! 💯 

Great work ! Thank you ! 🎉

lib/models/database/comment/comment_id_firestore.dart
```diff
@@ -1,8 +1,10 @@
 // The comment id is strong typed to avoid misuse
 import "package:flutter/foundation.dart";
+import "package:proxima/models/database/firestore/id_firestore.dart";
 
 @immutable
-class CommentIdFirestore {
+class CommentIdFirestore extends IdFirestore {
```
gruvw, line 6: I know I was not assigned to review this PR but I just noticed an important thing here:

We should avoid using inheritance here, we can use `implements` instead of `extend` and remove the constructor here:
https://github.com/ProximaEPFL/proxima/blob/6d50d25b2d9490c6b08fc8fa4c080fff917d1644/lib/models/database/firestore/id_firestore.dart#L4

test/mocks/services/mock_upvote_repository_service.dart
```diff
@@ -0,0 +1,42 @@
...
+    ParentIdFirestore? parentId,
+    UpvoteState? newState,
+  ) {
+    return super.noSuchMethod(
+      Invocation.method(
+        #setUpvoteState,
+        [userId, parentId, newState],
+      ),
+      returnValue: Future.value(),
+      returnValueForMissingStub: Future.value(),
```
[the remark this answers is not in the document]
yoannLafore: I think you are right :+1: 
Fix in ed3ecaaa262ac8b2d56272db7d02311dbdcb0942.

test/mocks/services/mock_upvote_repository_service.dart
yoannLafore: Good suggestion !
However I think we might want to keep it like this to be closer to the real prototype of the function which is:
```dart
Future<UpvoteState> getUpvoteState(
    UserIdFirestore userId,
    ParentIdFirestore parentId, {
    Transaction? transaction,
  })
```
What do you think ?

lib/models/database/comment/comment_id_firestore.dart
yoannLafore: You are right !
I wasn't aware of the `implements` in dart and thought that we only had `extends`.
Thanks for suggestion, this effectively makes more sense.
Fix in fb7379bb92f99a100b1115c0d4903d2072d41d42.

test/mocks/services/mock_upvote_repository_service.dart
CHOOSEIT: You are right 💯 that is a good idea.

lib/models/database/comment/comment_id_firestore.dart
gruvw: Dart has a lot of stuff when it comes to classes, even [mixins](https://dart.dev/language/mixins), be careful of the rabbit hole :)

lib/services/database/comment_upvote_repository_service.dart
```diff
@@ -0,0 +1,24 @@
+import "package:hooks_riverpod/hooks_riverpod.dart";
+import "package:proxima/models/database/comment/comment_data.dart";
+import "package:proxima/models/database/comment/comment_firestore.dart";
+import "package:proxima/models/database/comment/comment_id_firestore.dart";
+import "package:proxima/models/database/post/post_firestore.dart";
+import "package:proxima/models/database/post/post_id_firestore.dart";
+import "package:proxima/services/database/firestore_service.dart";
+import "package:proxima/services/database/upvote_repository_service.dart";
+
+final commentUpvoteRepositoryService = Provider.family<
```
JoachimFavre, line 10: I think this variable name should finish with `Provider`.

test/services/database/comment_upvote_repository_test.dart
```diff
@@ -0,0 +1,136 @@
...
+      firestore = FakeFirebaseFirestore();
+      postRepository = PostRepositoryService(firestore: firestore);
+      commentRepository = CommentRepositoryService(firestore: firestore);
+      commentDataGenerator = CommentDataGenerator();
+
+      final container = ProviderContainer(
+        overrides: [
+          firestoreProvider.overrideWithValue(firestore),
+          postRepositoryProvider.overrideWithValue(postRepository),
+          commentRepositoryProvider.overrideWithValue(commentRepository),
```
JoachimFavre, line 40: I don't think those overrides are necessary. The code still works on my machine without them (the firestore override is sufficient)
```suggestion
```

JoachimFavre (commented):
I have some very small comments, but overall this is a very good refactoring PR. The +300/-120 is frightening, but most of the code is either tests or moved.
Always a pleasure reading your PRs :)

lib/services/database/comment_upvote_repository_service.dart
yoannLafore: You are right !
Fix in 8f4c0e9ddc06351397b6f00fb0568b119f9f5a14.

test/services/database/comment_upvote_repository_test.dart
yoannLafore: Totally right !
Fix in ade3a650aa88f5ca6b9ef4a7c1933710db9a9460.

JoachimFavre (approved):
LGTM :)
Good job for this great PR!

yoannLafore:
Thanks a lot for all your suggestions !
I think the PR should be ready to go now :tada:

PR #185, opened 2024-04-30
User posts viewmodel

gruvw:
## User posts viewmodel

Fixes #132 

This PR adds the code for the profile page user posts viewmodel.

### Acceptance criteria

- [X] The user profile page "posts" tab should be filled with real user posts data (remove fake data).
- [X] The user should be able to refresh their post on their profile page using pull down motion (same as in overview).

### Result

See below that the acceptance criteria are met.

### Code

This PR adds the UI data class for a post on the profile page `UserPost`.

**Note**: as we recently created #184 we'll need to add the position to this data class at a later stage (TODO in code).

This PR also creates a new viewmodel for the user posts: `UserPostsViewModel`. It is consumed by the relate UI widget `ProfileUserPosts` which exposes the user posts to `ProfileInfoCard` that was already used to display fake data before.

Additionally, the `ProfileInfoColumn` has been slightly modified to incorporate an optional refreshable async `FutureVoidCallback` function called `onRefresh`.

**Note**: this PR is the foundation basis for the user post deletion capabilities that arise from #170 (another PR is in preparation for that).

#### Others

As I already worked a lot on the profile page + now the user posts UI and viewmodel I am starting to really know well this portion of the code base. Therefore, I think it would make sense that I also work (if nobody is opposed to it) on the future user comments viewmodel (once the model is ready) as they are related tasks and I should be pretty efficient at it now. Just something to keep it mind :)

PR #148, opened 2024-04-24
Logout button to profile page + profile refactor + loading button

gruvw:
Fixes #134 

This PR aims to move the logout button from its temporary position on the home page to the profile page.
In order to do, so I had to heavily refactor the Profile Page and apply a slight change to our error handling process.

**Acceptance criteria**:

- [X] No logout button on the home page
- [X] See a logout button on profile page
- [x] Clicking on the logout button on the profile page redirects to the login page without errors (because there was one at first).
- [X] The deletion button should be able to accept an arbitrary future as parameter and display a spinner while the deletion occurs.

**Result**: profile page app bar with logout button

image

see https://github.com/ProximaEPFL/proxima/pull/148#issuecomment-2076904196 for the loading deletion button.

**Note**: this PR also bumps the project version according to MS2 sprint 2.

**Note**: waiting for #144 PR to be merged.

PR #188, opened 2024-04-30
Delete user post from profile page

### LINKED ISSUE #170: Ability to delete posts on profile
Acceptance criteria:
- Can delete the posts from the list on the profile page

gruvw:
## Delete user post from profile page

Fixes #170 

**Note**: waiting on #185 to be merged first.

This PR adds the code required for the feature of allowing a user to delete a post they made, on their profile page.

### Acceptance criteria

- [X] Clicking on the delete button on a post card on the profile page deletes the post.
- [X] Clicking on the delete button on the past details popup also deletes the post.
- [X] The user posts displayed on their profile page is automagically refreshed when a post is deleted.
- [X] When a user doesn't have any post on their profile page, they are shown with a short helper message instead of _white void_ (empty list).

**Note**: using the icon spinner button that was added in #148 to provide feedback while the post gets deleted (see https://github.com/ProximaEPFL/proxima/pull/185#issue-2270841464).

### Result

See below that the acceptance criteria are met.

**Note** for reviewers: take a look at #185 first.


test/views/pages/profile/delete_post_profile_test.dart
```diff
@@ -0,0 +1,111 @@
...
+      final noPostHelper = find.text(ProfileUserPosts.noPostsInfoText);
+      expect(noPostHelper, findsOneWidget);
+
+      // Check no posts left in fake database
+      final userPosts =
+          await fakeFireStore.collection(PostFirestore.collectionName).get();
+      final posts =
+          userPosts.docs.map((data) => PostFirestore.fromDb(data)).toList();
+
+      expect(posts.isEmpty, true);
```
yoannLafore, line 72: I just think we can use the `isEmpty` matcher here:
```suggestion
      expect(posts, isEmpty);
```

test/views/pages/profile/delete_post_profile_test.dart
```diff
@@ -0,0 +1,111 @@
...
```
yoannLafore, line 108: Same here :+1: 
```suggestion
      expect(posts, isEmpty);
```

test/views/pages/profile/delete_post_profile_test.dart
```diff
@@ -0,0 +1,111 @@
...
+import "package:proxima/views/pages/profile/profile_data/profile_user_posts.dart";
+
+import "../../../mocks/data/firestore_post.dart";
+import "../../../mocks/data/firestore_user.dart";
+import "../../../mocks/data/geopoint.dart";
+import "../../../mocks/providers/provider_profile_page.dart";
+import "../../../mocks/services/setup_firebase_mocks.dart";
+
+void main() {
+  const delayNeededForAsyncFunctionExecution = Duration(seconds: 1);
```
yoannLafore, line 18: I was wondering if the `delayNeededForAsyncFunctionExecution` is truly necessary in the tests.
Because I tried to remove it and all the tests passed successfully thus I don't really know if we need them ?

yoannLafore (commented):
You did a great job at implementing the deletion of posts !
This was really a feature that was missing in our application and this will improve the user experience.

The code is very clean and looks good to me.
I have just made a few style suggestions below.

test/views/pages/profile/delete_post_profile_test.dart
gruvw: Yes, so my philosophy of whether to add this or not is basically when an async function is called by the viewmodel, I add the delay just to make sure we don't break future behavior that might arise.

We can remove them if you prefer to only have them where they are absolutely necessary, but I don't really see an argument why we should ?
Let me know what you prefer.

test/views/pages/profile/delete_post_profile_test.dart
gruvw: ok, fixed by 75ad921

test/views/pages/profile/delete_post_profile_test.dart
gruvw: ok, fixed by 75ad921

lib/viewmodels/user_posts_view_model.dart
```diff
@@ -37,6 +38,12 @@ class UserPostsViewModel extends AutoDisposeAsyncNotifier<UserPostsState> {
     return posts;
   }
 
+  Future<void> deletePost(PostIdFirestore postId) async {
```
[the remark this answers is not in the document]
gruvw: Absolutely right! Done in 33c9d4e

test/views/pages/profile/delete_post_profile_test.dart
Aderfish: I actually wonder what the idea behind this is too. Is the point to reduce the total timeout of the animation?Shouldn't you then do this instead.

```
await tester.pumpAndSettle(const Duration(milliseconds: 100), EnginePhase.sendSemanticsUpdate, delayNeededForAsyncFunctionExecution);
```

From what I understand the following instead increases the duration of the pump, which allows for less reactive widgets to pass.
```
await tester.pumpAndSettle(delayNeededForAsyncFunctionExecution);
```

Aderfish (changes requested):
Great PR! It's great to have this functionality in the app now. I just have the same question as @yoannLafore on what the delay in the tests is actually for.

test/views/pages/profile/delete_post_profile_test.dart
gruvw: In a similar way (and for the same reason), we always call `tester.pumpAndSettle` even though `test.pump` would work for many cases.

yoannLafore: Oh ok I see, thanks for your answer.
My worry is that I feel like it will make the tests take a longer time to pass if I understand correctly ?

Aderfish (approved):

test/views/pages/profile/delete_post_profile_test.dart
gruvw: I multiplied by 10 the number of seconds in `delayNeededForAsyncFunctionExecution` and it passed all the tests with 2 seconds less than with 1 second value on my machine (:rofl:).

Don't take this as a proper performance benchmark, of course, but I don't think it has a meaningful impact on the performances of the test suite.

gruvw: I also tried out of curiosity with `100 seconds` and I did not observe any difference in the timing measurements of the test suite.

yoannLafore: Ok, thank you for testing.
Then it's all good to me :+1:

yoannLafore (approved):
The PR looks good to me :tada:

test/views/pages/profile/delete_post_profile_test.dart
gruvw: See issue #201 for further discussion on the topic.

PR #189, opened 2024-04-30
Post creation and navigation refreshes home feed

### LINKED ISSUE #187: Refresh the overview page feed once the user created a post (navigation based refresh)

gruvw:
## Post creation and navigation refreshes home feed

Fixes #187 

This PRs adds the necessary code to automatically refresh the home feed of posts on navigation and on post creation.

### Acceptance criteria

- When navigating from another page back to the home feed, the posts on the feed are refreshed.
- After coming back from the post creation page the home feed is refreshed (we see our new post right away, no need to manually refresh).

### DIFF
--- a/test/mocks/overrides/override_home_view_model.dart
+++ b/test/mocks/overrides/override_home_view_model.dart
@@ -8,7 +8,7 @@ import "../data/post_overview.dart";
 /// This class is particularly useful for the UI tests where we want to expose
 /// particular data to the views.
 /// By default it exposes an empty list of [PostOverview] and does nothing on refresh.
-class MockHomeViewModel extends AsyncNotifier<List<PostOverview>>
+class MockHomeViewModel extends AutoDisposeAsyncNotifier<List<PostOverview>>
     implements HomeViewModel {
   final Future<List<PostOverview>> Function() _build;
   final Future<void> Function() _onRefresh;
--- a/lib/viewmodels/home_view_model.dart
+++ b/lib/viewmodels/home_view_model.dart
@@ -9,7 +9,7 @@ import "package:proxima/services/geolocation_service.dart";
 /// It fetches the posts from the database and returns a list of
 /// (postId: [PostIdFirestore], postOverview: [PostOverview]) objects to be displayed.
 /// These represent the overview data to be displayed associated to the corresponding post id.
-class HomeViewModel extends AsyncNotifier<List<PostOverview>> {
+class HomeViewModel extends AutoDisposeAsyncNotifier<List<PostOverview>> {
   HomeViewModel();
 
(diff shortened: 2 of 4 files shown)


CHOOSEIT (approved):
LGTM !

test/mocks/overrides/override_home_view_model.dart
```diff
@@ -8,7 +8,7 @@ import "../data/post_overview.dart";
 /// This class is particularly useful for the UI tests where we want to expose
 /// particular data to the views.
 /// By default it exposes an empty list of [PostOverview] and does nothing on refresh.
-class MockHomeViewModel extends AsyncNotifier<List<PostOverview>>
+class MockHomeViewModel extends AutoDisposeAsyncNotifier<List<PostOverview>>
```
Aderfish, line 11: This appears to work when I tested it by hand by changing points in the database. I suggest we add a test that does this. We can either wait for #193 to be merged to use challenges, or instead change the points value in the fake database before navigating back to the profile.

Aderfish (changes requested):
LGTM, I just think we should add a test for the profile getting updated as well.

Just as a comment, I'm not actually sure if this actually improves the user experience since the post feed refresh takes so long. The acceptance criteria are met though, and we can think about improving performance later. Maybe we will have to trade accuracy for that, or somehow refresh in the background.

test/mocks/overrides/override_home_view_model.dart
gruvw: Yes, absolutely.

The test was added in 6a0d070 :)

test/views/pages/profile/profile_page_test.dart
```diff
@@ -265,5 +264,33 @@ void main() {
...
+
+      // Change the user points
+      await userCollection
+          .doc(expectedUser.uid.value)
+          .set(expectedUser.data.withPointsAddition(10).toDbData());
+
+      // Navigate back to profile page
+      await tester.tap(find.byKey(AppTopBar.profilePictureKey));
+      await tester.pumpAndSettle(delayNeededForAsyncFunctionExecution);
+    });
```
Aderfish, line 294:
```suggestion

      // Check updated centauri points
      final updatedPoints = expectedUser.data.centauriPoints.toString();
      expect(find.textContaining(updatedPoints), findsOneWidget);
    });
```

I think you forgot to check the updated points.

gruvw: Right, mb thank you. It was in my local branch, but somehow I forgot to stage the changes...

I committed it in 1d59c7a. Please mark as resolve when it's ok for you.

Aderfish (approved):
LGTM now!

PR #190, opened 2024-04-30
Feed sorting buttons

### LINKED ISSUE #172: Allow the feed to be sorted according to the sort buttons
Acceptance criteria:
- general service for sorting posts
- view model for the feed, for the sorting buttons to work

JoachimFavre:
Fixes #172.

This PR adds a service for sorting posts, and uses it in the view-model of the feed to show the posts in the correct order. I have also had to add a very simple view-model for the chosen sorting option.

I think it is interesting for me to justify my design choices for the `PostSortingService`, since there was some debate on whether its sort function should take a general function or an enum. I chose the second, because I felt it allowed a better encapsulation and type safety. I did not find any example of a place where we would want a sorting order different from the ones in the feed. Moreover, in my case, I decided to give a score function as a parameter instead of a comparator. This makes, in my opinion, code more readable since we more clearly see the intended order. However, it requires a boolean to know if we want to sort by increasing or decreasing score. This needs to be bundled together with the score function, and the enum appears to be a good way for doing this as well. I hope you will agree with this design choice.

Furthermore, note that I had to do some modifications to the post and user generators for my testing. I could have done this in another PR, but I think we have already too many PRs for this sprint and that it would just add a blocker for this PR. I therefore decided to put those (very small) modifications in this PR.

Acceptance criteria:
- [x] general service for sorting posts
- [x] view model for the feed, for the sorting buttons to work


test/views/pages/home/dynamic_home_test.dart
```diff
@@ -0,0 +1,100 @@
...
+
+  group("Sort options", () {
+    testWidgets(
+      "Pushing the sort option button sets the correct order to the posts",
+      (tester) async {
+        await tester.pumpWidget(homepageWidget);
+        await tester.pumpAndSettle();
+
+        final element = tester.element(find.byType(HomePage));
+        final container = ProviderScope.containerOf(element);
```
CHOOSEIT, line 61: I feel that we can find a better way to get the providerScope and read results from the `feedSortOptionsProvider`. 

We can maybe take inspiration from @yoannLafore's repository testing, where it creates a `ProviderContainer` and reads the provider from it. ([ref](https://github.com/ProximaEPFL/proxima/pull/181/files#diff-9f98d592c3821b134d5ac116b3644a38358e76cd72a4b974d6468c572e7a04e3)).

I don't know if this the best way, but I found a way to make this idea working:
I create the homeWidget with: 
```dart 
homepageWidget = UncontrolledProviderScope(
  container: containerProvider,
  child: homePageApp,
);

 ```
where 
```dart
containerProvider = ProviderContainer(overrides: [
  firestoreProvider.overrideWithValue(firestore),
  geoLocationServiceProvider.overrideWithValue(geoLocationService),
]);

 ```
 
Using that, I am able to use: 
```dart 
expect(
  containerProvider.read(feedSortOptionsProvider),
  equals(sortOption),
);

// Expect the list to be in correct order
// We compare the post id values for an easier debugging.
final actualPostsOverview = await containerProvider.read(
  postOverviewProvider.future,
);


```

But since we never used `UncontrolledProviderScope` in our codebase, I am not sure if this is a better solution.

CHOOSEIT (commented):
Really clean. Great work 🎉.

I have a comment about a test.

However, overall, I really love the way you test your code, creating loops and functions to avoid code duplication. 💯

lib/services/sorting/post_sorting_service.dart
```diff
@@ -0,0 +1,59 @@
...
+  /// randomly shuffled that we sort by alphabetical order. If
+  /// putOnTop contains [A, D, K], then the reuslt will be
+  /// [A, D, b, c, e].
+  List<PostFirestore> sort(
+    List<PostFirestore> posts,
+    PostSortOption option,
+    GeoPoint position, {
+    Set<PostFirestore> putOnTop = const {},
+  }) {
+    int defaultComparator(a, b) {
```
yoannLafore, line 21: I just believe we can specify the parameter types in the function here and maybe rename them.
```suggestion
    int defaultComparator(PostFirestore postA, PostFirestore postB) {
```

lib/views/home_content/feed/feed_sort_option_chips.dart
```diff
@@ -0,0 +1,42 @@
+import "package:flutter/material.dart";
+import "package:hooks_riverpod/hooks_riverpod.dart";
+import "package:proxima/services/sorting/post_sort_option.dart";
+import "package:proxima/viewmodels/feed_sort_options_view_model.dart";
+
+/// This widget is the list of sort options for the timeline.
+class FeedSortOptionChips extends ConsumerWidget {
+  static final optionChipKey = Map.fromIterables(
```
yoannLafore, line 8: Very small suggestion, I think we can make this variable to the plurial given that it is a map with multiple keys.
```suggestion
  static final optionChipKeys = Map.fromIterables(
```

lib/views/home_content/feed/feed_sort_option_chips.dart
```diff
@@ -0,0 +1,42 @@
...
+        label: Text(sortOption.name),
+        onSelected: (value) {
+          if (value) {
+            ref
+                .read(feedSortOptionsProvider.notifier)
+                .setSortOption(sortOption);
+          }
+        },
+      );
+    }).toList();
```
yoannLafore, line 34: Good job with the UI :+1: 
Small suggestion, I believe that we can use the arrow synthax here.

```suggestion
        final sortOptions = PostSortOption.values
        .map(
          (sortOption) => ChoiceChip(
            key: optionChipKey[sortOption],
            selected: sortOption == selectedOption,
            label: Text(sortOption.name),
            onSelected: (value) {
              if (value) {
                ref
                    .read(feedSortOptionsProvider.notifier)
                    .setSortOption(sortOption);
              }
            },
          ),
        )
        .toList();
```

test/services/sorting/post_sorting_service_test.dart
```diff
@@ -0,0 +1,178 @@
...
+  final factor = ascending ? 1 : -1;
+  expect(iterable.isSorted((a, b) => factor * a.compareTo(b)), true);
+}
+
+void main() {
+  final sortingService = PostSortingService();
+
+  final postGenerator = FirestorePostGenerator();
+  final positions = GeoPointGenerator.generatePositions(userPosition0, 10, 0);
+  final posts = postGenerator.generatePostsAtDifferentLocations(positions);
```
yoannLafore, line 21: Very nice job with the testing ! 
I have little suggestion, I think we should declare those kind of variables as `late` that we initalize in the `setUp` function. I believe this is the convention we adopted for the other tests and it ensures that each test has those variables freshly created when running and so that they were not modified by other tests.

I also wonder if we could add a "global" group to encapsulate all the testing.
I think we could have something like:

```suggestion
  group("Testing post sorting service", () {
    late PostSortingService sortingService;

    late FirestorePostGenerator postGenerator;
    late List<GeoPoint> positions;
    late List<PostFirestore> posts;

    setUp(() {
      sortingService = PostSortingService();

      postGenerator = FirestorePostGenerator();
      positions = GeoPointGenerator.generatePositions(userPosition0, 10, 0);
      posts = postGenerator.generatePostsAtDifferentLocations(positions);
    });
    
    // All the other groups that are performing tests
    // ...
    
  }
```

Note: if you go with this suggestion, I think you'll need to also implement the next suggestion otherwise I believe you'll get an error because some variable won't be initialized.

test/services/sorting/post_sorting_service_test.dart
```diff
@@ -0,0 +1,178 @@
...
+  group("Nonempty onTop attribute", () {
+    final positionsToPutOnTop =
+        GeoPointGenerator.generatePositions(userPosition0, 5, 0);
+
+    /// The posts that we will ask to put on top, if in the list
+    final postsToPutOnTop =
+        postGenerator.generatePostsAtDifferentLocations(positionsToPutOnTop);
+
+    /// The posts that will actually be in the list to be sorted
+    final existingPostsToPutOnTop = postsToPutOnTop.take(3).toList();
```
yoannLafore, line 139: For similar reasons to my previous comment, I think we could make those `late` and initialize them in the `setUp`.
```suggestion
      late List<GeoPoint> positionsToPutOnTop;
      late List<PostFirestore> postsToPutOnTop;
      late List<PostFirestore> existingPostsToPutOnTop;

      setUp(() {
        positionsToPutOnTop =
            GeoPointGenerator.generatePositions(userPosition0, 5, 0);

        postsToPutOnTop = postGenerator
            .generatePostsAtDifferentLocations(positionsToPutOnTop);

        existingPostsToPutOnTop = postsToPutOnTop.take(3).toList();
      });
```

yoannLafore (commented):
You did a very good job at implementing the sorting service !
The code is very clean and robustly tested.

I also like how you made use of functions in your tests to reduce code duplication.
As @CHOOSEIT said, I also just think that it could have a little more documentation :+1: 

I have just made a few suggestions below.

Thank you for you work !
Great job :tada:

lib/services/sorting/post_sort_option.dart
```diff
@@ -0,0 +1,51 @@
...
+double voteScore(PostFirestore post, GeoPoint position) {
+  return post.data.voteScore.toDouble();
+}
+
+// Increase by 1 every day
+double dayScore(PostFirestore post, GeoPoint position) {
+  final timeInMs = post.data.publicationTime.millisecondsSinceEpoch;
+  final timeInDays = timeInMs / (1000.0 * 60.0 * 60.0 * 24);
+  return timeInDays;
+}
```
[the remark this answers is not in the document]
JoachimFavre: As I mention in my code, the absolute value of those score does not matter, only the relative values. This means that subtracting currentTime from every score does not change anything at best, and, at worst, makes it non-deterministic: `currentTime` will be evaluated every time we want to compare two posts, so, for two posts a,b published at the same time, we might have both a < b and b < a if it takes more than a millisecond to compute the score of the first one one.
Would you want me to increase  this documentation?

test/views/pages/home/dynamic_home_test.dart
JoachimFavre: By humble curiosity, why do you call this a "better way"? This looks like more boilerplate to me at first sight.

test/services/sorting/post_sorting_service_test.dart
JoachimFavre: I have a question and a remark on this suggestion:
- Does this really make sense to put all my tests in the same group? This really feels like using groups just for using groups, without having them giving real quality to me at first sight.
- My idea was that it was a lot of boilerplate to declare my variables as `late` and define them in the setup; since they never change. However, on reflection, this appears to be a good way to avoid future mistakes and non-deterministic tests. I'll implement this.

test/views/pages/home/dynamic_home_test.dart
CHOOSEIT: I feel like we are extracting the container from the UI, when we could have extracted it more globally without going through the UI, but I may be wrong.

test/services/sorting/post_sorting_service_test.dart
yoannLafore: - I don't know if it is strictly necessary but I quite like the idea having a big group that encapsulates everything and that tells what service we are testing. And then we have smaller groups to tell which part of the service we are testing. But this is only a personal opinion.
- Nice thank you :+1:

lib/services/sorting/post_sort_option.dart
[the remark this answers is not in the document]
yoannLafore: Oh ok I see thank you.
Adding a bit more documentation could be a good idea you're right :+1:

test/views/pages/home/dynamic_home_test.dart
JoachimFavre: What I did is what is recommended [in the documentation](https://riverpod.dev/docs/essentials/testing) for testing both widgets and providers at the same time. In my opinion, this is ok.

lib/services/sorting/post_sorting_service.dart
```diff
@@ -0,0 +1,59 @@
...
+import "package:proxima/services/sorting/post_sort_option.dart";
+
+class PostSortingService {
+  /// This method sorts the [posts] according to the [option]
+  /// and the [position] of the user. It also allows to put some
+  /// posts on top with the [putOnTop] set.
+  /// For instance, let's say that we have a list [A, b, c, D, e]
+  /// randomly shuffled that we sort by alphabetical order. If
+  /// putOnTop contains [A, D, K], then the reuslt will be
+  /// [A, D, b, c, e].
```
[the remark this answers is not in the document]
JoachimFavre: Well seen for the typo, fixed in 0f0e6f26c6958427a4f3032b1d8ae705846e37ef.

No `K` is not a typo, if an element is in `putOnTop` but not in the list, it will not be added to the list. This is for challenges later so that, if they are in the feed, they appear on top. I improved the documentation in c1909e7af60f5f2a37443cf91e365646da5052cb. Tell me if this is clearer.

lib/services/sorting/post_sorting_service.dart
```diff
@@ -0,0 +1,59 @@
...
+
+        return defaultComparator(a, b);
+      };
+    }
+
+    return posts.sorted(comparator);
+  }
+}
+
+final postSortingProvider = Provider<PostSortingService>(
```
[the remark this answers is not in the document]
JoachimFavre: Fixed in 4bc9f8a131b5635dbc7e72269dcdf31ce85847da.

lib/views/home_content/feed/feed_sort_option_chips.dart
```diff
@@ -0,0 +1,42 @@
...
+  Widget build(BuildContext context, WidgetRef ref) {
+    final selectedOption = ref.watch(feedSortOptionsProvider);
+
+    final sortOptions = PostSortOption.values.map((sortOption) {
+      return ChoiceChip(
+        key: optionChipKey[sortOption],
+        selected: sortOption == selectedOption,
+        label: Text(sortOption.name),
+        onSelected: (value) {
+          if (value) {
```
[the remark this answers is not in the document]
JoachimFavre: Fixed in 9d7384f7bcf8a380745db7b144ff99131f04d62c.

test/services/sorting/post_sorting_service_test.dart
```diff
@@ -0,0 +1,178 @@
...
+      positionDifference,
+      forcePositiveScore: true,
+    );
+
+    testOption(
+      PostSortOption.top,
+      "increases by 1 every upvote",
+      upvoteDifference,
+    );
+  });
```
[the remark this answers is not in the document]
JoachimFavre: Fix in bb5caded1305cb7d8977074298e655dc025ffacb. Tell me if this is clearer.

lib/services/sorting/post_sort_option.dart
```diff
@@ -0,0 +1,51 @@
...
+  /// It has a [name] for the ui, a [scoreFunction] to calculate the score of a post
+  /// when sorting it and a [sortIncreasing] to know if the sort should be done in
+  /// increasing order of score or not.
+  /// The exact value of the score does not matter, only their relative values. For
+  /// instance, adding 100'000 to all scores would not change anything.
+  const PostSortOption(this.name, this.scoreFunction, this.sortIncreasing);
+}
+
+// Increases by 1 every upvote, but decreases by 1 every day
+double hotScore(PostFirestore post, GeoPoint position) {
```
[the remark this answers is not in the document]
JoachimFavre: Fixed in 3f486895e8f6c6ad7f78b3bb60e3029aa494b14e.

lib/services/sorting/post_sort_option.dart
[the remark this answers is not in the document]
JoachimFavre: Indeed, some documentation was missing, I added it in 024c9bfc9b5d57e0fb1e6395b39f682e76a5c3c8.

In fact, writing documentation made me realise I had confused myself when writing this code and its corresponding test. This was fixed in d1ad77617eab46c9cdbad9f173d9cf632f321f56.

test/views/pages/home/dynamic_home_test.dart
CHOOSEIT: You are right. Thank you.

lib/services/sorting/post_sorting_service.dart
JoachimFavre: Fixed in d36c080e5f63c82ff79a2853138390e7856148d8.

lib/views/home_content/feed/feed_sort_option_chips.dart
JoachimFavre: Fixed in f5420ecd18367f611c0e3939ef52571ab692c501.

lib/views/home_content/feed/feed_sort_option_chips.dart
JoachimFavre: Fixed in b0dec4962087d323e1efed7f1238fd02a8cf7235.

test/services/sorting/post_sorting_service_test.dart
JoachimFavre: - If you're ok with this, I will not implement this suggestion. I think test names are already long enough, and that this is not strictly necessary.
- Fixed in d6c9c4ecd56bf2b95cdcdb3926ee37184c6624b8.

test/services/sorting/post_sorting_service_test.dart
JoachimFavre: Fixed in d6c9c4ecd56bf2b95cdcdb3926ee37184c6624b8.

test/services/sorting/post_sorting_service_test.dart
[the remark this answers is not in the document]
CHOOSEIT: Much clearer; thank you !

test/services/sorting/post_sorting_service_test.dart
yoannLafore: I understand, it's alright for me :+1:

test/services/sorting/post_sorting_service_test.dart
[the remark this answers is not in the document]
CHOOSEIT: Much clearer for the `testOption` function, thank you. However, I think that the inner mechanism of the function needs a little more explanation. Like, why `referenceScore` and what it represents, because this function is really abstract. :smile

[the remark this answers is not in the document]
JoachimFavre: Oh, I understand better. Is 1ffcb65fdd81fcb829228ecd467bf2591ba02662 good for you?

yoannLafore (approved):
The PR looks good to me now.
Amazing job :tada:

test/services/sorting/post_sorting_service_test.dart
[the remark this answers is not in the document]
CHOOSEIT: Much clearer 💯 . Thank you.

CHOOSEIT (approved):
LGTM now ! 🎉 

Great work 💯

JoachimFavre:
I found some mistake in my code. I'll let @yoannLafore re-review before I merge.

yoannLafore (approved):
Good catch !
Thank you for the change, this looks good to me :+1:

PR #191, opened 2024-04-30
Mock code generation

AlbertsReisons:
Just a small PR that started because I wanted to test out Mockito's code generation.

I think using it will not be difficult, and could save us a lot of time and boilerplate code. I have modified the code for the `UserRepositoryService` and `PostUpvoteRepositoryService` to use it.

**Usage**: It is possible to generate code locally once by running
`dart run build_runner run`

It is also possible to start a background process that will automatically update generated files when necessary by running
`dart run build_runner serve --live-reload`

I recommend the second option, and have added it in `scripts/generate_code` so that it is easy to find.

**Git**: From what I understand, it is better not to commit generated source files (more info [here](https://pub.dev/packages/build_runner#source-control)). I have therefore added a `*.mocks.dart` exception in the `.gitignore`, and added source generation in the test workflow so that it can still run. The other workflows are not affected as they target only the application code, not the tests.

**Mockito**: Writing mocks can now be done simply with 4 lines of code:

```
import "package:mockito/annotations.dart";
import "package:proxima/services/database/user_repository_service.dart";

@GenerateNiceMocks([MockSpec<UserRepositoryService>()])
void empty() {}
```

The last line is because the compiler complains if it cannot find any declaration in the file. I agree that it is ugly and I welcome propositions (the analyzer even complains if I decide to make it private).
Then if the name of the current file is `a.dart`, the mock will be generated in the same package in `a.mocks.dart`, and we can than import that file in our tests.


JoachimFavre:
Pushed to the next sprint, see #194.

camillelnne (approved):
LGTM :)
Please fix conflicts before merging

JoachimFavre:
We decided that you would fix the conflicts @camillelnne 🐘

PR #192, opened 2024-04-30
Architecture Diagram Milestone 2

CHOOSEIT:
This PR aims to update the architecture diagram for the milestone 2.

Proposed Architecture diagram (Updated with comments): 
Proxima Arichtecture Diagram
Because life's too short for monochrome—yes, the colors were necessary! 😄

Acceptance criteria:
- [x] The architecture diagram in the README must contain the new architecutre


CHOOSEIT:
I updated the diagram, ensuring 90° arrow line angles, and removed a random text floating around (thanks to @TheTexanCodeur)

gruvw:
Good job on updating this diagram! The additional colors were indeed necessary and I think they give a nice visual help to quickly know the first class services attached to any UI component.

I also noted that you took the diagram a few steps further than the actual project state, but I think it is ok and will simplify our life in the future :)

Maybe we could add a note for every service, as now it looks a bit strange to only have upvote and sign in that are not explained.

They don't need to be lengthy, but a short helping text on each service sounds like a good addition to me.

Other than that, the updated diagram looks good to me :rocket:

gruvw (approved):
LGTM now :)

yoannLafore (commented):
Good job, the diagram looks beautiful :tada: 

I only have the following small suggestion:
Given that #181 implements the voting system for the comments, I think we can add an arrow "UpVote Repository" -> "Comment Repository".
Or otherwise we can remove the arrow "UpVote Repository" -> "Posts Repository" given that the upvote repository is made more general in #181.
I think both choices are fine depending on the interpretation.
What do you think ?

Other than that, everything looks good to me :+1: 
Great job !

CHOOSEIT:
> I only have the following small suggestion: Given that #181 implements the voting system for the comments, I think we can add an arrow "UpVote Repository" -> "Comment Repository". Or otherwise we can remove the arrow "UpVote Repository" -> "Posts Repository" given that the upvote repository is made more general in #181. I think both choices are fine depending on the interpretation. What do you think ?


You are completely right. I think that the first option would be better to really understand their interaction even if we abstract the repository in code. 👍🏼

yoannLafore (approved):
Thanks for the modification
The PR looks good to me.
Thank you for your great work :tada:

PR #215, opened 2024-05-03
Fix challenge repository crash when no available challenge

### LINKED ISSUE #216: Fix the challenge page crashing when there is no available challenge
Acceptance criteria:
- the challenge page must not crash when there is no challenge

JoachimFavre:
Fixes #216.

This fixes a crash from the challenge when there is no available challenge. It crashes [here](https://github.com/firebase/flutterfire/blob/953bf929bf19e7bbb3564c69901f5a4fca5fc981/packages/cloud_firestore/cloud_firestore/lib/src/query.dart#L734) in the real firestore. This cannot be tested with our mock firestore, since [this behaviour is not emulated in it](https://github.com/atn832/fake_cloud_firestore/blob/46fd92164bbc543534c3fdd7fe390c535bbd2847/lib/src/mock_query.dart#L230). In fact, we already had a test that checked the case where there is no challenge in-range, `challenge_repository_test.dart`:"Challenge range works"; so no regression test can be done.

Acceptance criteria:
- [x] the challenge page must not crash when there is no challenge

### DIFF
--- a/lib/views/home_content/challenge/challenge_list.dart
+++ b/lib/views/home_content/challenge/challenge_list.dart
@@ -12,7 +12,7 @@ class ChallengeList extends ConsumerWidget {
     final asyncChallenges = ref.watch(challengeProvider);
 
     Widget emptyChallenge = const Center(
-      child: Text("No challenges available here"),
+      child: Text("No challenge available here!"),
     );
 
     return CircularValue(
--- a/lib/views/home_content/challenge/challenge_list.dart
+++ b/lib/views/home_content/challenge/challenge_list.dart
@@ -11,16 +11,22 @@ class ChallengeList extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     final asyncChallenges = ref.watch(challengeProvider);
 
+    Widget emptyChallenge = const Center(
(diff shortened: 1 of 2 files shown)


lib/views/home_content/challenge/challenge_list.dart
```diff
@@ -11,16 +11,22 @@ class ChallengeList extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     final asyncChallenges = ref.watch(challengeProvider);
 
+    Widget emptyChallenge = const Center(
+      child: Text("No challenges available here"),
```
JoachimFavre, line 15: Small typo ;)
```suggestion
      child: Text("No challenges available here"),
```

CHOOSEIT: You are completely right ! Fixed in: [3aef58c](https://github.com/ProximaEPFL/proxima/pull/215/commits/3aef58ca7c73026cd665adc21e76f62341567fef)

gruvw (approved):

CHOOSEIT (approved):
Looks good to me 💯 . Thank you !

PR #222, opened 2024-05-04
Display a congratulatory message when winning points

### LINKED ISSUE #211: Congratulatory message on challenge completeted
Acceptance criteria:
- Display the number of points earned
- Small animation (maybe, if simple and pretty)
- Update the penpot accordingly

JoachimFavre:
Fixes #211.

This adds a small snack bar when the user gets points after finishing a challenge.
<p align="center">

</p>

Acceptance criteria:
- [x] Display the number of points earned
- [x] Small animation (maybe, if simple and pretty)  [I did not find any sober animation that is interesting]
- [ ] Update the penpot accordingly  [doing this soon, once I know how to handle this with the links getting stale]

### DIFF
--- a/lib/utils/ui/centauri_snack_bar.dart
+++ b/lib/utils/ui/centauri_snack_bar.dart
@@ -1,0 +1,19 @@
+import "package:flutter/material.dart";
+
+/// the state ScaffoldMessenger.of(context)
+/// Show a snackbar telling the user they won [value] Centauri points.
+/// [scaffoldMessengerState] is the current state of the scaffolds,
+/// on which to display the snackbar. It can typially be found
+/// using [ScaffoldMessenger.of(context)].
+void showCentauriPointsSnackBar(
+  int value,
+  ScaffoldMessengerState scaffoldMessengerState,
+) {
+  scaffoldMessengerState.showSnackBar(
+    SnackBar(
+      content: Text(
+        "You won $value Centauri! 🎉",
+      ),
+    ),
+  );
+}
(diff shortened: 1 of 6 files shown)


lib/utils/ui/centauri_snack_bar.dart
```diff
@@ -0,0 +1,23 @@
...
+  int value,
+  ScaffoldMessengerState scaffoldMessengerState,
+) {
+  scaffoldMessengerState.showSnackBar(
+    SnackBar(
+      behavior: SnackBarBehavior.floating,
+      duration: centauriPointsSnackBarDuration,
+      content: Text(
+        "You won $value Centauri! 🎉",
+      ),
```
Aderfish, line 20:
```suggestion
      ),
      showCloseIcon: true,
```

I suggest we add a close icon to this snack bar, it will make the message easier to dismiss once it is read.

Aderfish (commented):
The code looks good to me. I suggest that we add a closing icon on the snack, but we can discuss that.
Otherwise I'm happy to approve once we have the penpot updated.

lib/utils/ui/centauri_snack_bar.dart
JoachimFavre: I'm not sure I agree, a close icon is not very pretty in my opinion. I'm not sure it is a big issue if the user cannot close it immediately, since this is a rare pop-up.

I can decrease the time this snackbar appears however, if you wish. Currently, it stays for 4 seconds, which is the default value.

Aderfish: I understand, I didn't consider that this would be a rare popup since I was testing it over and over again. Let's leave it like this.

Aderfish (approved):
The mockup looks good, this is ready to merge.

Issue #212, opened 2024-05-03
Create the viewmodel for displaying comments


camillelnne:
Acceptance criteria:
- The comments of post are displayed under the post in the post page
- Use already implemented UI
- The comments are sorted from newest to oldest
- Test comments content.

PR #225, opened 2024-05-05
Comment viewmodel

yoannLafore:
Hello,
This PR aims to provide the viewmodel to display the comments of a post on the post page (issue #212).
This will allow the use to see the comments that have been posted under a post.

Concretely, this PR does the following :

**View model for displaying comments:**
This PR provides a view model for exposing the comments under a post in a newest -> oldest fashion. This is done through the class `CommentViewModel` and more specifically with the provider `commentListProvider`.
This provider takes as argument a `PostIdFirestore` which represent the id of post for which we want to get the comments.

The `CommentViewModel` also implements a `refresh` method that allows to re-fetch the comments of the post and exposing them in the state.

**Use the view model in the UI:**
The `PostPage` widget will now use the `commentListProvider` to display the comments under the posts.
It will also call `CommentViewModel.refresh` on scrolling the page down to  refresh the list of comments.

**Mock viewmodel:**
To allow for compatibility with the tests, a class `MockCommentViewModel` has been created to have a mock implementation of the comment view model that can be used in the static tests. Also, the necessary providers have been modified to take this `MockCommentViewModel` as parameter.

**Tests for the viewmodel:**
Tests for the `CommentViewModel` have been been added.

#### Live testing
You can live test this PR functionnality by the following:
Set the emulator location to be in the [Vortex](https://www.google.com/maps/place/Vortex/@46.5246451,6.5750268,18.82z/data=!4m6!3m5!1s0x478c31250537c2ed:0xf9160729430c8b8b!8m2!3d46.5244962!4d6.575197!16s%2Fg%2F11h__kzrvz?entry=ttu).
You should see a post named "Singe", I have manually added comments under this post.
You can click on it and you should see the following:

#### Acceptance criteria:
- [x] The comments of post are displayed under the post in the post page
- [x] Use already implemented UI
- [x] The comments are sorted from newest to oldest
- [x] Test comments content.


lib/views/pages/post/post_page.dart
```diff
@@ -38,17 +39,22 @@ class PostPage extends HookConsumerWidget {
...
       CompletePostWidget(
         key: completePostWidgetKey,
         post: postOverview,
       ),
       const SizedBox(height: 10),
-      CommentList(
-        key: commentListWidgetKey,
-        comments: comments,
-      ),
+      commentList,
```
JoachimFavre, line 57: This a very personal opinion, but I feel like creating a lot of the interface using different variables makes it less readable. Personally, I'd prefer (quite a lot) if this variable was inlined.

test/mocks/overrides/override_comment_view_model.dart
```diff
@@ -0,0 +1,51 @@
...
+
+  MockCommentViewModel({
+    Future<List<CommentPost>> Function(PostIdFirestore arg)? build,
+    Future<void> Function()? onRefresh,
+  })  : _build =
+            build ?? ((PostIdFirestore arg) async => List<CommentPost>.empty()),
+        _onRefresh = onRefresh ?? (() async {});
+
+  @override
+  Future<List<CommentPost>> build(PostIdFirestore arg) => _build(arg);
```
JoachimFavre, line 24: I'm also not sure of the name `arg` here and line 13.

test/mocks/overrides/override_comment_view_model.dart
```diff
@@ -0,0 +1,51 @@
...
+
+final mockNonEmptyCommentViewModelOverride = [
+  commentListProvider.overrideWith(
+    () => MockCommentViewModel(
+      build: (PostIdFirestore arg) async => testComments,
+    ),
+  ),
+];
+
+final mockLoadingCommentViewModelOverride = [
```
JoachimFavre, line 44: Is it normal this varialbe is used nowhere?

test/viewmodels/comment_view_model_integration_test.dart
```diff
@@ -0,0 +1,149 @@
...
+          () async {
+        const numberOwners = 3;
+        const numberComments = 19;
+
+        // Create and add the comments owner to the database
+        final owners =
+            FirestoreUserGenerator.generateUserFirestore(numberOwners);
+        for (final owner in owners) {
+          await userRepository.setUser(owner.uid, owner.data);
+        }
```
JoachimFavre, line 68: I think that this can be extracted to a helper function, like the function `addPosts` of the `PostDataGenerator`.

test/viewmodels/comment_view_model_integration_test.dart
```diff
@@ -0,0 +1,149 @@
...
+        final actualComments =
+            await container.read(commentViewModelProvider.future);
+
+        // Check that all the comments are present
+        expect(actualComments, unorderedEquals(expectedComments));
+
+        final expectedSortedComments = expectedComments.sorted(
+          (commentA, commentB) =>
+              -commentA.publicationDate.compareTo(commentB.publicationDate),
+        );
```
JoachimFavre, line 108: I'm not a big fan of this test, since this is just a re-implementation of your code. I think that a more interesting test would be to check the intuitive order, i.e. that, when iterating over all posts, the post at index i should have been published after the one at index i+1.

test/viewmodels/comment_view_model_integration_test.dart
```diff
@@ -0,0 +1,149 @@
...
+        final owner = FirestoreUserGenerator.generateUserFirestore(1).first;
+        await userRepository.setUser(owner.uid, owner.data);
+
+        final commentData =
+            commentDataGenerator.createMockCommentData(ownerId: owner.uid);
+        await commentRepository.addComment(
+          postId,
+          commentData,
+        );
+
```
JoachimFavre, line 132: I think verifying here that there is no comment is also interesting. I'm not entierly sure that the verification on lines 118-120 is really that useful, but this is your implementation choice to move it here or copy it here.

JoachimFavre (commented):
This is very good code, which is well tested, great job! :)

Just for sake of completeness, I want to mirror here a discussion we had offline. In my opinion, comments should be sorted in the other order since, currently, discussions are in  reverse order. A vote was issued on our group chat, and it appeared the majority agreed with Yoann's implementation. Their main argument is that this allows to see what is currently said on a post, without having to scroll back to the end.

lib/viewmodels/comment_view_model.dart
```diff
@@ -0,0 +1,61 @@
...
+import "package:proxima/services/database/comment_repository_service.dart";
+import "package:proxima/services/database/user_repository_service.dart";
+
+/// This view model is used to fetch the comments of a post.
+/// It fetches the comments under the post with the id [arg] and returns
+/// a list of [CommentPost] objects to be displayed.
+class CommentViewModel
+    extends AutoDisposeFamilyAsyncNotifier<List<CommentPost>, PostIdFirestore> {
+  @override
+  Future<List<CommentPost>> build(PostIdFirestore arg) async {
```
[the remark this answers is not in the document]
yoannLafore: I completely agree with you on that.
However, doing so yields a warning from the dart formatting because this is the parameter of an overriden method. The warning is the following:

```dart
The parameter name 'postId' doesn't match the name 'arg' in the overridden method.
Try changing the name to 'arg'.dart[avoid_renaming_method_parameters](https://dart.dev/lints/avoid_renaming_method_parameters)
```

Thus I need to keep the parameter named `arg`.

lib/views/pages/post/post_page.dart
yoannLafore: Yes I think you are right.
I had done that because `CommentList` is now wrapped up with a `CircularValue` and I though it might little to much to have everything inline.
But looking again at it I think it is more readable when inlined.

Fix in e75f07f46da9b8bd15600d7737a3215b790d2aba.

test/mocks/overrides/override_comment_view_model.dart
yoannLafore: Yes, that was to expose the main mocks that someone might be interested in when designing a future test.
But I can remove it if you think it would be better like that ?

test/viewmodels/comment_view_model_integration_test.dart
yoannLafore: You are totally right.
Actually I found out that we already have a function `setUsersFirestore` that I can directly use here.
Thanks for pointing that out :+1: 

Fix in 08ea0b5b67a76bc271fda01bbe6a7c94e49f0249.

test/viewmodels/comment_view_model_integration_test.dart
yoannLafore: That is a good point indeed :+1: 
I have refactored the test to check the intuitive ordering in ca7e18d983cb1757ab20c83df6bc5cca88fa8021.
Thanks for the suggestion !

test/viewmodels/comment_view_model_integration_test.dart
yoannLafore: Yes this is a very good point :+1: 
I have moved the check accordingly.
Note that I still need to call `await container.read(commentViewModelProvider.future);` a first time at the beginning of the test to force a first call to the `build()` method of the viewmodel that will expose the empty state.

Then I check that that this state was not updated by adding a comment to the repository.

Implemented in a778d3a188395f4a689740ae75ab74fe4ec50ee0.

test/mocks/overrides/override_comment_view_model.dart
yoannLafore: I don't seem to get a warning here when renaming the parameter (I believe that is because this is a test file).
However I think that for consistency we might want to keep it named `arg` if we need to keep the parameter of `CommentViewModel.build` named `arg`.

lib/viewmodels/comment_view_model.dart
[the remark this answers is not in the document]
JoachimFavre: There is maybe a way to suppress this warning locally, would it make sense? What do you think?

test/mocks/overrides/override_comment_view_model.dart
JoachimFavre: I'm not sure I see in which context it would be needed. Though, it would allow to potentially avoid a merge conflict and improve structure if someone needs it at some point. 
I personally think this piece of code is unnecessary and that unused code is not good, but I'll leave this choice at your discretion.

lib/viewmodels/comment_view_model.dart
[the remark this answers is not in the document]
yoannLafore: We can indeed remove the warning by adding `// ignore: avoid_renaming_method_parameters`.
However, I am not sure that this is really a good practice to do so.

I personnaly prefer to avoid ignoring the warning and to keep the parameter named `arg`.
Also, I realised that the class exposes `arg` as an attribute that we can access anywhere in the class methods. And we cannot rename this one.

test/mocks/overrides/override_comment_view_model.dart
yoannLafore: After giving another though about I also didn't found a good context for it's usage.
Thus I'll follow your suggestion :+1: 

Fix in 5124caddd0411991f6d889fe6d019106f2317f5c.

lib/viewmodels/comment_view_model.dart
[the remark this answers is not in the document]
JoachimFavre: Yes, I agree ignoring warnings is a dangerous habit. Let's keep this as you did! :)
Do you think a small comment justifying this variable name would make sense?

test/viewmodels/comment_view_model_integration_test.dart
JoachimFavre: This is not exactly what I meant, I thought of something closer to:
https://github.com/ProximaEPFL/proxima/blob/22e9a498134fbdcaec99dce2e5e1c95045eb3996/test/mocks/data/firestore_post.dart#L90C1-L99C1

However, I'm realising a function for two lines of code is a bit overkill. I'd therefore understand you stay like this, this is already much better! :)

lib/viewmodels/comment_view_model.dart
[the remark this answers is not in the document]
yoannLafore: Yes, that's a good idea !
Done in d9114e92bda04e2000dbdbb9633fbffb071504d5.

test/viewmodels/comment_view_model_integration_test.dart
yoannLafore: Oh ok I see, that's actually a good idea and could be useful in the future.
I implement this in 8ec59b4b04e97d79fb9142785fe8c0c8401fcafb.

Thanks for the suggestion :)

JoachimFavre: Interestingly, in `PostDataGenerator`, `addPosts` is not static method: it is outside of the class. I think that `addComments` should not be inside the class out of consistency, but I'd understand if you want to keep this as is since this is, in my opinion, more coherent.
I moreover think it is missing a bit of documentation ;)

yoannLafore: I believe that `addPosts` is indeed located inside `FirestorePostGenerator`.
Thus I think that I will let `addUsers` inside `FirestoreUserGenerator` if that's ok for you.

I have also added a line of documentation :)
Done in 03b0684ec430e869d91b964108de1a1c51e83df9.

JoachimFavre: Yes indeed, my bad! Thank you :)

JoachimFavre (approved):
This LGTM now, good job! :)

yoannLafore:
Thank you for all your great suggestions !
I think the codebase is now ready to be merged :tada:

Issue #224, opened 2024-05-05
Create viewmodel for adding comments


yoannLafore:
Acceptance criteria:
- The user is able to add a comment on a post
- When the user adds a comments, a refresh happens such that the user directly sees its comments under the post
- The user cannot post a comment that is blank
- If the user tries to post a blank comment, the input field displays an error telling the user that he cannot post a blank comment

PR #226, opened 2024-05-05
Add comment viewmodel

yoannLafore:
Hello,
This PR aims to provide the view model to allow the user to comment under a post (issue #224).
With this PR, the user is able to comment under posts.

Concretely, this PR does the following:

**Model for state of adding a comment:**
This PR adds the class `NewCommentState` which represents the current state in the addition of a new comment. It exposes a possible error message and if the comment has been posted or not.

**Viewmodel for adding comment:**
This PR adds a view model for adding comment under a post. This is done through the class `NewCommentViewModel` inspired from the implementation of `NewPostViewModel`.
This `NewCommentViewModel` takes a parameter `PostIdFirestore` that represents the post under which the comment is added and it exposes the state `NewCommentState`.

One can a comment using the method `NewCommentViewModel.tryAddComment`. 
This one will validate the comment (check for emptyness). If the comment is invalid, an error message state is exposed.
Otherwise, the addition of the comment continues normally.

**Link the view model to the UI:**
The implementation of `BottomBarAddComment` has been modified to use the real `NewCommentViewModel`.
This required a little refactoring of the class to keep a good implementation.

**Tests for the view model:**
This PR adds some tests for both the viewmodel in itself as well as the view using the view model.

#### Live testing
You can live test these PR features as follows:
Open the app and click on a post.
Then you can try to add a comment under the post and you should see it appear immediately after clicking the add button.
You can also try to add an empty comment and you should see the text field displaying an error.

You should get the following behavior when adding valid comment: 

#### Acceptance criteria:
- [ ] The user is able to add a comment on a post
- [ ] When the user adds a comment, a refresh happens such that the user directly sees its comments under the post
- [ ] The user cannot post a comment that is empty
- [ ] If the user tries to post a empty comment, the input field displays an error telling the user that he cannot post a empty comment

**Important note:** Waiting for #225 to be merged to change the base branch to main and merge.

### DIFF
--- a/lib/models/ui/new_comment_state.dart
+++ b/lib/models/ui/new_comment_state.dart
@@ -1,0 +1,24 @@
+/// Represent the state of the new comment form
+/// It contains the content error message (null if there is no error)
+/// and a flag to indicate if the comment was posted
+class NewCommentState {
+  final String? contentError;
+  final bool posted;
+
+  NewCommentState({
+    required this.contentError,
+    required this.posted,
+  });
+
+  @override
+  bool operator ==(Object other) {
+    return other is NewCommentState &&
+        other.contentError == contentError &&
+        other.posted == posted;
+  }
+
+  @override
+  int get hashCode {
+    return Object.hash(contentError, posted);
+  }
+}
--- a/lib/viewmodels/new_comment_view_model.dart
+++ b/lib/viewmodels/new_comment_view_model.dart
@@ -12,7 +12,7 @@ import "package:proxima/viewmodels/login_view_model.dart";
 /// post id [PostIdFirestore] is provided as an argument.
 class NewCommentViewModel
     extends AutoDisposeFamilyAsyncNotifier<NewCommentState, PostIdFirestore> {
-  static const String _contentError = "Please fill out your comment";
+  static const String contentEmptyError = "Please fill out your comment";
 
   @override
   Future<NewCommentState> build(PostIdFirestore arg) async {
@@ -26,7 +26,7 @@ class NewCommentViewModel
     if (content.isEmpty) {
       state = AsyncData(
         NewCommentState(
-          contentError: _contentError,
+          contentError: contentEmptyError,
           posted: false,
         ),
       );
--- a/lib/viewmodels/new_comment_view_model.dart
+++ b/lib/viewmodels/new_comment_view_model.dart
@@ -77,3 +77,8 @@ class NewCommentViewModel
     return state.value!;
   }
 }
+
+final newCommentStateProvider = AsyncNotifierProvider.autoDispose
+    .family<NewCommentViewModel, NewCommentState, PostIdFirestore>(
+  () => NewCommentViewModel(),
+);
--- a/lib/viewmodels/new_comment_view_model.dart
+++ b/lib/viewmodels/new_comment_view_model.dart
@@ -8,6 +8,8 @@ import "package:proxima/models/ui/new_comment_state.dart";
 import "package:proxima/services/database/comment_repository_service.dart";
 import "package:proxima/viewmodels/login_view_model.dart";
 
+/// The view model for adding a new comment to a post whose
+/// post id [PostIdFirestore] is provided as an argument.
 class NewCommentViewModel
     extends AutoDisposeFamilyAsyncNotifier<NewCommentState, PostIdFirestore> {
   static const String _contentError = "Please enter a coment";
--- a/lib/viewmodels/new_comment_view_model.dart
+++ b/lib/viewmodels/new_comment_view_model.dart
@@ -12,7 +12,7 @@ import "package:proxima/viewmodels/login_view_model.dart";
 /// post id [PostIdFirestore] is provided as an argument.
 class NewCommentViewModel
     extends AutoDisposeFamilyAsyncNotifier<NewCommentState, PostIdFirestore> {
-  static const String _contentError = "Please enter a coment";
+  static const String _contentError = "Please fill out your comment";
 
   @override
   Future<NewCommentState> build(PostIdFirestore arg) async {
--- a/lib/viewmodels/new_comment_view_model.dart
+++ b/lib/viewmodels/new_comment_view_model.dart
@@ -37,6 +37,12 @@ class NewCommentViewModel
     return true;
   }
 
+  /// Resets the state to be ready for adding a new comment.
+  Future<void> reset() async {
+    ref.invalidateSelf();
+    await future;
+  }
+
   /// Verifies that the content is not empty, then adds a new comment to the database.
   /// If the content is empty, the state is updated with the appropriate error message.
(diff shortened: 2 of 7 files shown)


lib/models/ui/new_comment_state.dart
```diff
@@ -0,0 +1,24 @@
...
+  final String? contentError;
+  final bool posted;
+
+  NewCommentState({
+    required this.contentError,
+    required this.posted,
+  });
+
+  @override
+  bool operator ==(Object other) {
```
gruvw, line 14: nice you remembered to follow our decision about `identical` :)

lib/views/pages/post/post_page_widget/bottom_bar_add_comment.dart
```diff
@@ -1,59 +1,105 @@
...
+    final contentController = useTextEditingController();
+
+    final asyncNewCommentState = ref.watch(
+      newCommentStateProvider(parentPostId),
+    );
+
+    return CircularValue(
+      value: asyncNewCommentState,
+      builder: (context, newCommentState) {
+        // The avatar of the current user on the left
```
gruvw, line 37: It would make sense to create an additional `StatelessWidget` to abstract the content of the circular value here. The builder method is way too long and hard to read imo.
We could abstract it in two or three widgets that we could place in `views/pages/post/components`.

lib/views/pages/post/post_page.dart
```diff
@@ -83,10 +83,11 @@ class PostPage extends HookConsumerWidget {
           padding:
               // This is necessary to prevent the keyboard from covering the bottom bar
               EdgeInsets.only(bottom: MediaQuery.of(context).viewInsets.bottom),
-          child: const BottomBarAddComment(
+          child: BottomBarAddComment(
             key: bottomBarAddCommentKey,
             //TODO: Replace with actual username
             currentDisplayName: "Username",
```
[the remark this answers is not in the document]
yoannLafore: I agree with you.
However I believe that @CHOOSEIT is already working on it in #227 thus I prefer to avoid making modifications in this PR to reduce merge conflicts.

lib/views/pages/post/post_page_widget/bottom_bar_add_comment.dart
```diff
@@ -1,59 +1,105 @@
...
+                FocusManager.instance.primaryFocus?.unfocus();
+              }
             },
           ),
-        ),
-      ],
+        );
+
+        return Row(
+          // Align items to the start of the cross axis
```
gruvw, line 94: I think we should avoid comments that paraphrase the code. We can read the same information on the next line. (more on the picky side I admit)

test/viewmodels/new_comment_view_model_integration_test.dart
```diff
@@ -0,0 +1,162 @@
...
+      newCommentViewModelProvider = newCommentStateProvider(postId);
+    });
+
+    test("Add a valid comment and reset correctly", () async {
+      final validContent = commentDataGenerator.createMockCommentData().content;
+
+      final newCommentViewModel =
+          container.read(newCommentViewModelProvider.notifier);
+
+      //  --- Check the state before adding the comment
```
gruvw, line 60: What is this notation with three dashes ? `// ---`

Do we use this somewhere else in testing ? I don't remember seeing it and don't know what it means.

lib/viewmodels/new_comment_view_model.dart
```diff
@@ -0,0 +1,94 @@
...
+    final postId = arg;
+
+    final commentData = CommentData(
+      content: content,
+      ownerId: currentUserId,
+      publicationTime: Timestamp.now(),
+      voteScore: 0,
+    );
+
+    final commentRepository = ref.read(commentRepositoryProvider);
```
gruvw, line 77: It's good practice to stack all `ref.` actions at the top of the method.

lib/viewmodels/new_comment_view_model.dart
```diff
@@ -0,0 +1,94 @@
...
+import "package:proxima/models/database/comment/comment_data.dart";
+import "package:proxima/models/database/post/post_id_firestore.dart";
+import "package:proxima/models/ui/new_comment_state.dart";
+import "package:proxima/services/database/comment_repository_service.dart";
+import "package:proxima/viewmodels/login_view_model.dart";
+
+/// The view model for adding a new comment to a post whose
+/// post id [PostIdFirestore] is provided as an argument.
+class NewCommentViewModel
+    extends AutoDisposeFamilyAsyncNotifier<NewCommentState, PostIdFirestore> {
```
gruvw, line 14: I don't think we want this notifier to be `AutoDispose` as if you make a small mistake and accidentally press "back" while typing a comment, you won't be able to get it back.

I think a simple solution would be that we keep the user's text on the comment field.
Otherwise, we could also have a small pop up asking for confirmation to leave the post page when the comment field is not empty.
I am more in favor of the first option, tell me what you think.

gruvw (changes requested):
Great implementation :+1: I left a few remarks, but the code is good overall and the acceptance criteria are fulfilled.

test/viewmodels/new_comment_view_model_integration_test.dart
yoannLafore: This was to have a more visual separation in the test code.
But I am realising that I didn't used it elsewhere.
Thus I remove it to be consistent with the rest of the codebase,

Done in d235ffe3ed2da2db32217cd8652335ab4fdfd112.

lib/viewmodels/new_comment_view_model.dart
yoannLafore: That is a good point :+1: 

Fix in 1991a150a71d284949ea4e5d32ba40167f26e55b.

lib/viewmodels/new_comment_view_model.dart
yoannLafore: You are totally right, thank you for pointing it out :+1: 

I chose to go with the first option which I find to be the less intrusive.
To do so, I removed the `AutoDispose` from the notifier and I moved the text field controller to be kept inside the viewmodel like that it is not destroyed when navigating back.

I implement this in 315d1e665f6bbf3f161dc814c63f37e57aab9abe.

lib/views/pages/post/post_page_widget/bottom_bar_add_comment.dart
yoannLafore: You are right, the circular value is not very readable.
I'll do my best to provide a good refactor.
So, I have created 3 new widgets to separate the UI elements of the bottom bar.
And I have putted them under `views/pages/post/components/comment`

I do that in b948c795a3e24729d2311df6b1ea1b05c05c490b.

gruvw:
Good job, all your fixes looks ok for me :)

gruvw (approved):

PR #228, opened 2024-05-06
Challenge logout error

### LINKED ISSUE #217: Bug: error message on login out from challenges
How to reproduce: from Challenges page, go to profile page and logout

Image

Aderfish:
Fixes #217 

To fix this I created a new uid provider that throws an exception instead of returning null, `validUidProvider`

I also modified the circular value to ignore exceptions that contain the debug text instead of just starting with it. This is because I was getting text of the type "Exception: DEBUG user is not logged in". 

I modified previous parts of the code to use that provider as well.

I also fixed a mistake in the error text of the profile viewmodel, which is unrelated but I saw it while refactoring.

Acceptance criteria:
- [x] error popup no longer shows when logging out after viewing challenges and then profile


test/end2end/app_test.dart
```diff
@@ -84,6 +90,30 @@ void main() {
     await createPost(tester);
     await deletePost(tester);
   });
+
+  testWidgets("Logout from challenges does not cause error", (tester) async {
```
JoachimFavre, line 94: I'm not sure this test really has its place in this file. This is not an end-to-end test, or part of one.

lib/viewmodels/login_view_model.dart
```diff
@@ -21,13 +22,28 @@ final isUserLoggedInProvider = Provider<bool>((ref) {
   return ref.watch(userProvider).valueOrNull != null;
 });
 
-/// Firebase logged in user id provider
+/// Firebase logged in user id provider, returns null if the user is not logged
+/// in
```
JoachimFavre, line 26: I think adding some small warning can be interesting:
```suggestion
/// in. [validUidProvider] should almost always be used: its
/// error does not cause a pop-up to be shown by the circular
/// value, which is typically what one want since the user will
/// get navigated back to the log-in page anyway.
```

JoachimFavre (commented):
This is very good code, thank you for doing a regression test! :)

I have some comments hereinbelow, which should not take you too long to fix.

Moreover, out of humble curiosity, why did you choose not to user `validUidProvider` on line 98 of `create_account_view_model.dart` ?
https://github.com/ProximaEPFL/proxima/blob/234f6178110835fa1a093d645cbd958f6ddea3be/lib/viewmodels/create_account_view_model.dart#L98

Aderfish:
You're right, I think we can add it to the account creation view model too. At first glance and since it is linked with account creation, I thought there was some business logic there. Actually it was just ignoring the problem silently. I tested creating a new account by hand and it still works.

test/end2end/app_test.dart
Aderfish: I refactored the e2e test folder, you can have a look.

JoachimFavre: I'm not sure I get your refactor. This is not an end-to-end test, is this? Moreover, is this refactor really meaningful in this PR? Shouldn't this be another PR?

Aderfish: Ok I changed the test to be just on the challenge view-model. I left some positive modifications to the end to end test that I made on the way, but it is just a small refactor that is necessary anyway if we want to add other tests. I think we can keep this in this pr.

JoachimFavre: Yeah, I'm not sure it has entierely its place in this PR, but I guess this is fine since this is already done and probably had to be done at some point. This is much better, thank you :)

JoachimFavre (approved):
This LGTM now, good job fixing this! :)

I love how most of the modifications of this PR are just tabulations in `challenge_view_model_integration_test.dart`.

Issue #206, opened 2024-05-02
Delete the subcollection of a post when a poost is deleted


yoannLafore:
Currently, when a post is deleted, only the post document is deleted but not its subcollections `voters` and `comments`.
Acceptance criteria:
- When a post is deleted, all its subcollections are deleted recursively

PR #229, opened 2024-05-06
Recursive post deletion

Aderfish:
This PR addresses #206.

It ensures that all the subcollections of a post are deleted on post deletion. Right now, this includes comments, upvotes and upvotes of comments. This is done by adding `deleteAll` methods to the comments and upvotes repository. From what I have found, it is not possible to access all the subcollections of a document in flutter, [more info](https://firebase.google.com/docs/firestore/query-data/get-data#list_subcollections_of_a_document). This means that this solution is not generic and will have to be adapted with specific subcollection names and more helper functions if the structure of the database evolves.
 We are also not using cloud functions, which could have made this easier.

Acceptance criteria:
- [x] When a post is deleted, all its subcollections are deleted recursively


lib/services/database/post_repository_service.dart
```diff
@@ -31,7 +43,13 @@ class PostRepositoryService {
 
   /// This method deletes the post with id [postId] from the database
```
JoachimFavre, line 44: I think it is worht mentionning it also removes all its subcollection:
```suggestion
  /// This method deletes the post with id [postId] from the database.
  /// This means removing the corresponding document, and all the
  /// subcollections.
```

test/services/database/comment_repository_test.dart
```diff
@@ -248,6 +248,26 @@ void main() {
 
         expect(post.data.commentCount, equals(5));
       });
+
+      test("can delete all comments", () async {
```
JoachimFavre, line 252: You are not testing that the comment upvotes are deleted too (which should be a failing test here I think). To do this, you will need acess to the comment upvote repository service, so I'd suggest the following refactor: you can create a function outputting the repository service corresponding to some post id in `comment_upvote_repository_service.dart`, and use this function in the provider. You can then use this function in this test.
I recommend you contacting @yoannLafore to discuss this, as he may also implement need/implement this for his tasks.

Edit: After some reflection, I am realising that, instead of a method creating the `UpvoteRepository` it might be interesting to simply create a new class (to better match the other repository interface):
```
class CommentUpvoteRepositoryService
    extends UpvoteRepositoryService<CommentIdFirestore> {
  CommentUpvoteRepositoryService({
    required super.firestore,
    required PostIdFirestore postId,
  }) : super(
          parentCollection: firestore
              .collection(PostFirestore.collectionName)
              .doc(postId.value)
              .collection(CommentFirestore.subCollectionName),
          voteScoreField: CommentData.voteScoreField,
        );
}

final commentUpvoteRepositoryProvider = Provider.family<
    UpvoteRepositoryService<CommentIdFirestore>,
    PostIdFirestore>((ref, postId) {
  return CommentUpvoteRepositoryService(
    firestore: ref.watch(firestoreProvider),
    postId: postId,
  );
});
```
Some documentation would probably need to be added to this class, naturally.

test/services/database/post_repository_test.dart
```diff
@@ -47,6 +54,43 @@ void main() {
       expect(actualPost.exists, false);
     });
 
+    test("Delete deletes subcollections", () async {
```
JoachimFavre, line 57: I think the test name can be slightly improved:
```suggestion
    test("Delete method deletes subcollections", () async {
```

test/services/database/post_repository_test.dart
```diff
@@ -47,6 +54,43 @@ void main() {
...
+
+      final commentRepository = CommentRepositoryService(firestore: firestore);
+      await CommentFirestoreGenerator()
+          .addComments(4, post.id, commentRepository);
+
+      final upvoteRepository = UpvoteRepositoryService<PostIdFirestore>(
+        firestore: firestore,
+        parentCollection: firestore.collection(PostFirestore.collectionName),
+        voteScoreField: VoteFirestore.votersSubCollectionName,
+      );
```
JoachimFavre, line 73: I think a `PostUpvoteRepsoitoryService` class would also be interesting. See my previous comment for all the details of my reasoning.

JoachimFavre (commented):
Great code, as always! :)

I just have some comment below.

Moreover, I spotted what I think is a small error: you don't delete the comment upvote when deleting a comment.

Furthermore, I'm feeling the `deleteComment` and `deleteAllComments` share quite a bit of code. I however don't really find a way to improve this easily. What do you think?

lib/services/database/post_repository_service.dart
Aderfish: 8098704de70971a2826a70ea6e0b88afc73ee568

test/services/database/post_repository_test.dart
Aderfish: 8098704de70971a2826a70ea6e0b88afc73ee568

test/services/database/comment_repository_test.dart
Aderfish: The specialized providers exist, I don't think its for this pr to decide whether we need specialized classes as well.

I added a test to the single comment deletion that checks that upvotes are deleted as well. 30561afd5e78ed75a66439b524df03c611a968e3

Aderfish:
You're right, I factorized the common code and made sure single comment deletion deletes upvotes as well. Let me know what you think.

test/services/database/comment_repository_test.dart
Aderfish: After more discussion, with @yoannLafore as well, we decided this was a good idea but it was easier to do by making the current constructor of `UpvoteRepositoryService` private and make simpler public constructors that do not rely on the database structure. This was done in #237.

I integrated this refactor into this issue in 5a43590d69a679cbfa0db0e06800e3da40470ddf

test/services/database/post_repository_test.dart
Aderfish: See above discussion.

lib/services/database/comment_repository_service.dart
```diff
@@ -88,27 +89,84 @@ class CommentRepositoryService {
...
+      parentPostId,
+    );
+
+    for (final comment in comments.docs) {
+      await _deleteCommentNoCountUpdate(
+        parentPostId,
+        CommentIdFirestore(value: comment.id),
+        batch,
+        commentUpvoteRepository,
+        false,
```
JoachimFavre, line 136: I would prefer this to be a named parameter. I think it would improve the code readability in a way that is significant enough for the time taken to change it. What do you think?

test/services/database/comment_repository_test.dart
JoachimFavre: Thank you :)

JoachimFavre (changes requested):
The code was of great quality, now it is military-grade. Good job :)
I have a last comment left, it should not take you long to fix. Apart from that, LGTM :D

JoachimFavre (approved):

lib/services/database/comment_repository_service.dart
Aderfish: c31fa834d53b1f3b1c8fb5cef0ffdcebb582c335

PR #234, opened 2024-05-08
Time conversion service

gruvw:


lib/services/human_time_service.dart
```diff
@@ -0,0 +1,58 @@
...
+  String textTimeSince(DateTime dateTime) {
+    final referenceClock = currentDateTimeCallback();
+
+    final textTime = timeago.format(
+      dateTime,
+      locale: _relativeLocale,
+      clock: referenceClock,
+    );
+
+    if (_agoExceptions.contains(textTime)) {
```
TheTexanCodeur, line 33: Here the contains method only checks one corner case ("now"). Could we use a String "_agoException"  instead of a List<String> "_agoExceptions". In this case we would just check if "_agoException" is equal to "textTime".

If other exceptions will be tested in the future, then I understand the current implementation.

gruvw: Yes maybe we will discover other special cases (like "now") in the future. This is why I implemented it like this.

TheTexanCodeur (commented):
Thanks for you work!

The PR is clean, the code clear and well stuctured.

I just made a small comment but overall this PR looks ready in my opinion :)

lib/services/human_time_service.dart
TheTexanCodeur: Ok so we keep it like that!

TheTexanCodeur (approved):
LGTM!

PR #235, opened 2024-05-08
Test human readable time and distances values in UI

### LINKED ISSUE #209: Test posts and challenges distance
For now, only the proper display of the distance is tested, not their value. Same for the dates

Acceptance criteria:
- Test the value of distance of Challenges and posts
- Test the value of the date of posts and comments

gruvw:
Fixes #209

### Timing and distances values testing

This PR's goal is to test two new parts of the application:

- The distances values between posts displayed as human readable format.
- The new human readable values provided by the new service added in PR #234.

The testing relies heavily on Dependency Injection to make sure relative times are consistent across different testing environments and time zones.

Don't hesitate to ask question below if anything in the code is not clear :)

**Note**: first relevant commit is 1a62d3f.

#### Code

The testing is composed of two types of tests:

- Unit tests for the human time service that are done in `test/services/human_time_service_test.dart`
- Component testing for the human time service and the distances in the UI testing files.

I also added the providers override and helpers for testing the human time service:

- With constant time callback functions definitions + the two testing providers `constantDateTimeCallbackProvider` and `constantHumanTimeServiceProvider` that go with it: `test/mocks/providers/provider_human_time_service.dart`
- The simple `humanTimeServiceOverride` overriding the corresponding provider: `test/mocks/overrides/override_human_time.dart`

I think I used our new testing code correctly (after refactor in #122), if I did something that goes against this refactor's philosophy please let me know :)

#### Acceptance criteria (from #209)

- [x] Test the value of distance of Challenges and posts
- [x] Test the value of the date of posts and comments

**Note**: this PR also moves the `mockito` package to dev dependencies.

### DIFF
--- a/test/mocks/providers/provider_post_page.dart
+++ b/test/mocks/providers/provider_post_page.dart
@@ -1,6 +1,7 @@
 import "package:flutter/material.dart";
 import "package:hooks_riverpod/hooks_riverpod.dart";
 import "package:proxima/models/database/user/user_id_firestore.dart";
+import "package:proxima/models/ui/post_overview.dart";
 import "package:proxima/services/database/comment_repository_service.dart";
 import "package:proxima/viewmodels/login_view_model.dart";
 import "package:proxima/views/navigation/routes.dart";
@@ -42,4 +43,20 @@ ProviderScope postPageProvider(
     ],
     child: postPage,
   );
+}
+
+ProviderScope customPostOverviewPage(PostOverview post) {
+  final customPostApp = MaterialApp(
+    onGenerateRoute: generateRoute,
+    home: PostPage(
+      postOverview: post,
+    ),
+  );
+
+  final app = ProviderScope(
+    overrides: mockEmptyCommentViewModelOverride,
+    child: customPostApp,
+  );
+
+  return app;
 }
--- a/test/mocks/providers/provider_post_page.dart
+++ b/test/mocks/providers/provider_post_page.dart
@@ -62,7 +62,7 @@ ProviderScope customPostOverviewPage(PostOverview post) {
     ),
   );
 
-  final app = ProviderScope(
+  return ProviderScope(
     overrides: [
       ...mockDynamicUserAvatarViewModelTestLoginUserOverride,
       ...mockEmptyCommentViewModelOverride,
@@ -70,6 +70,4 @@ ProviderScope customPostOverviewPage(PostOverview post) {
     ],
     child: customPostApp,
   );
-
-  return app;
 }
--- a/test/mocks/providers/provider_post_page.dart
+++ b/test/mocks/providers/provider_post_page.dart
@@ -11,6 +11,7 @@ import "../data/post_overview.dart";
 import "../overrides/override_comment_view_model.dart";
 import "../overrides/override_firestore.dart";
 import "../services/mock_comment_repository_service.dart";
+import "provider_human_time_service.dart";
 
 // Create a post page with the first post from the testPosts list
 final postPage = MaterialApp(
@@ -54,7 +55,10 @@ ProviderScope customPostOverviewPage(PostOverview post) {
   );
 
   final app = ProviderScope(
-    overrides: mockEmptyCommentViewModelOverride,
+    overrides: [
+      ...mockEmptyCommentViewModelOverride,
+      constantHumanTimeServiceProvider,
+    ],
     child: customPostApp,
   );
 
--- a/test/mocks/providers/provider_post_page.dart
+++ b/test/mocks/providers/provider_post_page.dart
@@ -13,7 +13,6 @@ import "../overrides/override_dynamic_user_avatar_view_model.dart";
 import "../overrides/override_firestore.dart";
 import "../overrides/override_human_time.dart";
 import "../services/mock_comment_repository_service.dart";
-import "provider_human_time_service.dart";
 
 // Create a post page with the first post from the testPosts list
 final postPage = MaterialApp(
@@ -65,6 +64,7 @@ ProviderScope customPostOverviewPage(PostOverview post) {
 
   final app = ProviderScope(
     overrides: [
+      ...mockDynamicUserAvatarViewModelTestLoginUserOverride,
       ...mockEmptyCommentViewModelOverride,
       ...humanTimeServiceOverride,
     ],
--- a/test/mocks/providers/provider_post_page.dart
+++ b/test/mocks/providers/provider_post_page.dart
@@ -10,6 +10,7 @@ import "package:proxima/views/pages/post/post_page.dart";
 import "../data/post_overview.dart";
 import "../overrides/override_comment_view_model.dart";
 import "../overrides/override_firestore.dart";
+import "../overrides/override_human_time.dart";
 import "../services/mock_comment_repository_service.dart";
 import "provider_human_time_service.dart";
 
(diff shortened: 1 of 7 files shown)


test/mocks/providers/provider_post_page.dart
```diff
@@ -51,3 +53,23 @@ ProviderScope postPageProvider(
...
+
+ProviderScope customPostOverviewPage(PostOverview post) {
+  final customPostApp = MaterialApp(
+    onGenerateRoute: generateRoute,
+    home: PostPage(
+      postOverview: post,
+    ),
+  );
+
+  final app = ProviderScope(
```
TheTexanCodeur, line 65: Here could we directly return the ProviderScope in order to stay consistent with the definition of "postPageProvider" above?

test/services/human_time_service_test.dart
```diff
@@ -0,0 +1,116 @@
...
+      // Check that the specific absolut time value is correct
+      expect(
+        actualComplexHumanText,
+        expectedComplexHumanText,
+      );
+    });
+
+    test("Check special case 'now' relative time", () {
+      // Setup actual relative values for 'now' case
+      final nowTime = constantTestingTime.add(const Duration(seconds: 1));
```
TheTexanCodeur, line 60: I think that we should substract the time here, because it's weird to check how much time passed since a date in the future.

test/services/human_time_service_test.dart
```diff
@@ -0,0 +1,116 @@
...
+
+      const expectedNowRelativeTime = "now";
+
+      // Use the service to get relative human time
+      final actualNowTimeHumanText = humanTimeService.textTimeSince(
+        nowTime,
```
TheTexanCodeur, line 66: In my opinion, "nowTime" can be a confusing variable name, because it's suggest that it's the actual time (something similar to DateTime.now).  Do you think we should rename it (something like "justBeforeTime")?

test/views/pages/post/post_page_values_test.dart
```diff
@@ -0,0 +1,107 @@
...
+import "package:proxima/models/database/post/post_id_firestore.dart";
+import "package:proxima/models/ui/post_overview.dart";
+import "package:proxima/views/pages/post/post_page_widget/complete_post_widget.dart";
+
+import "../../../mocks/data/post_overview.dart";
+import "../../../mocks/providers/provider_post_page.dart";
+import "../../../mocks/services/setup_firebase_mocks.dart";
+
+void main() {
+  // Custom post for testing specific date and distances
```
TheTexanCodeur, line 13: Even if it's easier to understand the test with the custom post declared here, isn't it better if this mock data is declared in the file`test/mocks/data/post_overview.dart`

test/services/human_time_service_test.dart
```diff
@@ -0,0 +1,116 @@
...
+      final time1 = DateTime.utc(2002);
+
+      const expectedSimpleHumanText = "Tuesday, January 1, 2002 00:00";
+
+      // Use the service to get human time
+      final actualSimpleHumanText = humanTimeService.textTimeAbsolute(
+        time1,
+      );
+
+      // Check that the absolut time value is correct
```
TheTexanCodeur, line 28: Just a minor typo.

```suggestion
      // Check that the absolute time value is correct
```

test/services/human_time_service_test.dart
```diff
@@ -0,0 +1,116 @@
...
+      );
+
+      const expectedComplexHumanText = "Tuesday, November 21, 2023 01:00";
+
+      // Use the service to get human time
+      final actualComplexHumanText = humanTimeService.textTimeAbsolute(
+        time2,
+      );
+
+      // Check that the specific absolut time value is correct
```
TheTexanCodeur, line 51: Same
```suggestion
      // Check that the specific absolute time value is correct
```

test/services/human_time_service_test.dart
```diff
@@ -0,0 +1,116 @@
...
+
+      // Check that the time value is correctly 'now'
+      expect(
+        actualNowTimeHumanText,
+        expectedNowRelativeTime,
+      );
+    });
+
+    test("Check 10 minutes relative time", () {
+      // Setup specific time values for 10 min before contsant time
```
TheTexanCodeur, line 77: Just a minor typo.

```suggestion
      // Setup specific time values for 10 min before constant time
```

test/services/human_time_service_test.dart
```diff
@@ -0,0 +1,116 @@
...
+
+      // Check that the time value is correct for 10 minutes
+      expect(
+        actualMinuteTimeHumanText,
+        expectedRelative10Minutes,
+      );
+    });
+
+    test("Check 4 days relative time", () {
+      // Setup specific time values for 4 days before contsant time
```
TheTexanCodeur, line 97: Same

```suggestion
      // Setup specific time values for 4 days before constant time
```

TheTexanCodeur (commented):
Good job!

I think that those new tests are easy to understand thanks to the structure of the code and the comments. I made a few comments, but this PR looks ready for the most part.

test/services/human_time_service_test.dart
gruvw: ty, done in 688736d

test/services/human_time_service_test.dart
gruvw: ty, done in 688736d

test/services/human_time_service_test.dart
gruvw: ty, done in 688736d

test/services/human_time_service_test.dart
gruvw: ty, done in 688736d

test/views/pages/post/post_page_values_test.dart
gruvw: I don't think this post data will be useful to anyone else as it is closely coupled with the current test and I placed it here to avoid polluting the global testing namespace.

That being said, to stay consistent with the other mock data I applied your suggestion in 144bea8

test/services/human_time_service_test.dart
gruvw: Ok, done in 9507b49

test/services/human_time_service_test.dart
gruvw: Yes, it does not matter, as the `timeago` library does check this and displays now instead. Also, pretty much all the default posts that we have in all tests will end up being in the future, as the base fixed date that I chose is in the year `2000` so in the past. In the end I could place the fixed date in the future to avoid this issue, but I does not change anything as at some point it will end up being in the past anyway so why not set it in the past directly :)

I still fixed your suggestion in c0be566

test/mocks/providers/provider_post_page.dart
gruvw: I don't really like the way it was done above, so this is why I did it differently, but I didn't want to refactor it because it was not related to this PR...

Anyway, as it really is only a detail, I still applied your comment in aa0a976.

gruvw:
Ok @TheTexanCodeur I applied all your requested changes. Please let me know if there is anything else to change in this PR.

test/services/human_time_service_test.dart
TheTexanCodeur: Thanks for the clarifications!

TheTexanCodeur (approved):
Thank you for taking the suggestions into account ! The PR LGTM now 👍🏼

Issue #239, opened 2024-05-10
Bug fix: challenges query fails when more than 30 posts


camillelnne:
Acceptance criteria: 
- Challenge getter does not throw an error when more than 30 posts are available to become challenges.

<img src="https://github.com/ProximaEPFL/proxima/assets/101193378/6968ed80-f264-4a13-937f-e0dbfc38918d" width="200">

PR #261, opened 2024-05-11
Fix challenge repository crash when more than 30 available posts.

CHOOSEIT:
Fixes: #239 

This PR aims to fix a crash from the challenge repository whenever the number of available posts to be used as a challenge is greater than or equal to 30. 

This PR adds a regression test for this bug.

Acceptance criteria:
- [x] Challenge getter does not throw an error when more than 30 posts are available to become challenges.


lib/services/database/challenge_repository_service.dart
```diff
@@ -146,20 +146,25 @@ class ChallengeRepositoryService {
...
-        .where(FieldPath.documentId, whereIn: possiblePostsStringIds)
-        .get();
+    final alreadyDonePostsSnap = await pastChallengesCollectionRef.get();
 
+    // The [whereIn] argument of the where method crashed
+    // if the query has more than 30 elements for the real firestore.
+    // So we do the filtering locally rather than using the whereIn parameter
+    // of the where method.
     final alreadyDonePosts = alreadyDonePostsSnap.docs
+        .where((donePost) => possiblePostsStringIds.contains(donePost.id))
```
gruvw, line 167: Here I think we should avoid string comparison of IDs. It defeats the purposes of wrapping IDs in opaque types. (I am not really for wrapping IDs in the first place but as it is the decision when went with I think it is important to stick to it.)

Therefore, we should probably go with something like:

```dart
final alreadyDonePosts = alreadyDonePostsSnap.docs
    .map((post) => PostIdFirestore(value: post.id))
    .where((donePost) => possiblePosts.contains(donePost))
    .toSet();
```

That way, we don't accidentally make a mistake and compare two string IDs that do not reference the same types of objects (in theory).

On top of that, it is a bit more concise, avoids repetition and the need for the intermediary `possiblePostsStringIds` mapping iteration altogether.

test/services/database/challenge_repository_test.dart
```diff
@@ -128,6 +128,39 @@ void main() {
...
+        await firestore
+            .collection(UserFirestore.collectionName)
+            .doc(uid.value)
+            .collection(ChallengeFirestore.pastChallengesSubCollectionName)
+            .doc(pastPostId.value)
+            .set(data.toDbData());
+
+        //Create 30 posts and get the challenges
+        FirestorePostGenerator generator = FirestorePostGenerator();
+        var challenges = await challengeRepository.getChallenges(uid, userPos);
```
gruvw, line 154: I am not sure why the use of `var` was chosen here, as `challenges` is not used before its final reassignment (using the exact same expression).

I think I would just make more sense like this:

```dart
final generator = FirestorePostGenerator();
await generator.addPostsReturnDataOnly(firestore, inChallengeRange, 30);

final challenges =
    await challengeRepository.getChallenges(uid, userPos);
```

Maybe there is a side effect that I am not aware of that would make my suggestion not behave identically ?

gruvw (changes requested):
Good job on this bug fix! The code is well documented, and it is a great idea to add a regression test for this :+1:

test/services/database/challenge_repository_test.dart
```diff
@@ -128,6 +128,39 @@ void main() {
       expect(challenges.length, 1);
       expect(challenges.first.postId, posts[1].id); // the other post
     });
+
+    group("Regression tests", () {
+      //It seems that if you don't have any pastChallenges the mocked firestore
```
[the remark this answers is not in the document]
CHOOSEIT: You are completely right 💯 ! Thank you.

Added in: [2df2009](https://github.com/ProximaEPFL/proxima/pull/261/commits/2df20098f85f7da351217effbca9ff0ada28ff78)

lib/services/database/challenge_repository_service.dart
CHOOSEIT: Good catch, you are completely right.

Fixed in: [56d1c86](https://github.com/ProximaEPFL/proxima/pull/261/commits/56d1c861c20f432d4b0e85eb50053e4ae16e5786)

Thank you.

test/services/database/challenge_repository_test.dart
CHOOSEIT: You are completely right; getting two times the challenges of the user is not useful, as the first time none will be returned. Moreover it does not have any side effect that could affect the test.

Fixed in: [671e885](https://github.com/ProximaEPFL/proxima/pull/261/commits/671e8850f2640b10537e089b8443cc704b90d3ef)

Thank you 🚀

test/services/database/challenge_repository_test.dart
```diff
@@ -128,6 +128,40 @@ void main() {
...
       expect(challenges.first.postId, posts[1].id); // the other post
     });
+
+    group("Regression tests", () {
+      // It seems that if you don't have any pastChallenges the mocked firestore
+      // will don't do any check on the query, so we need to add a pastChallenge to
+      // the user.
+      test("Challenge getter when more than 30 available posts works",
+          () async {
+        //Create a post that will be used as a completed past challenge.
```
gruvw, line 138: I meant to apply https://github.com/ProximaEPFL/proxima/pull/261#discussion_r1597566590 consistently everywhere in the file (and for future code too).

CHOOSEIT: You are completely right, I thought it was the only one and didn't double checked. 

I now double checked in both files, I think it should be good in [0ad5856](https://github.com/ProximaEPFL/proxima/pull/261/commits/0ad585698b97273f0daea96f1090d22f22398f39).

Thank you !

gruvw (approved):

gruvw:
@CHOOSEIT can we delete this branch, or is there a reason why we kept it ?

CHOOSEIT:
We can may bad @gruvw 😃

PR #262, opened 2024-05-12
Apply dart code metrics suggestions

TheTexanCodeur:
This small PR applies the dart code metrics suggestions. It is important to evaluate from time to time the changes that this metrics tool proposes, especially with the fact that we don't have sonar cloud.

Note that there are still some suggestions left unaddressed that I judged as irrelevant. Don't hesitate to discuss about those that you think should be taken into account.

"File name does not match with first enum name." : I didn't take this suggestion into account because the enum name wouldn't make sense anymore.


camillelnne (commented):
Thanks for doing this refactor work!

Please note that the correct spelling is "refactor" and not "refractor". Typos might cause issues when computing the changelog from the conventional commits.

Could you please explain why you didn't apply some of the suggestions? 
Especially for the ones like " Expected a sync function but got async." that you did fix in some files but not others. All suggestions marked by a warning icon in the summary seemed necessary to me so I am curious about your reasoning :)

lib/services/database/upvote_repository_service.dart
```diff
@@ -86,13 +87,13 @@ class UpvoteRepositoryService<ParentIdFirestore extends IdFirestore> {
         ? await transaction.get(voteStateCollection)
         : await voteStateCollection.get();
 
-    if (!voteState.exists) {
-      return VoteState.none;
-    } else {
+    VoteState getActualState() {
```
gruvw, line 90: Why is this a local function instead of a final variable ?

lib/services/sorting/post/post_score_functions.dart
```diff
@@ -4,16 +4,16 @@ import "package:proxima/models/database/post/post_firestore.dart";
...
   PostFirestore post,
-  GeoPoint position,
+  GeoPoint userPosition,
 );
 
 /// Increases by 1 every upvote, but decreases by 1 every day
-double hotScore(PostFirestore post, GeoPoint userPosition) {
-  return voteScore(post, userPosition) - dayScore(post, userPosition);
+double hotScore(PostFirestore post, GeoPoint _) {
+  return voteScore(post, _) - dayScore(post, _);
```
gruvw, line 12: The `_` notation is only useful for unused variable. Here the argument is actually used, so I think we should keep the original argument name.

lib/services/database/upvote_repository_service.dart
gruvw: I think it is because otherwise `voteState.data()` might be null and therefore causing a null pointer exception on the `!` call right ?

If this is the reason why I think some documentation should be added to better understand this.

gruvw: Or maybe we could keep the original guard clause on this:

```dart
if (!voteState.exists) {
    return VoteState.none;
}
```

and simply remove the `else` clause.

I think that way the code should be more readable, and I don't think DCM would complain.

lib/services/sorting/post/post_sorting_service.dart
```diff
@@ -25,31 +25,27 @@ class PostSortingService {
...
-          return 1;
-        }
-
-        return defaultComparator(a, b);
-      };
-    }
+
+    putOnTop.isEmpty
+        ? comparator = defaultComparator
+        : comparator = (a, b) {
```
gruvw, line 36: The assignment should be done outside of the ternary operator to avoid repetition: `comparator = ... ? ... : ...`

We might also want a local private function definition instead of inlining it like that (for the else clause).

lib/services/sorting/post/post_score_functions.dart
```diff
@@ -22,15 +22,17 @@ double voteScore(PostFirestore post, GeoPoint userPosition) {
 /// it to -postDate since the same constant added to all scores (currentDate)
 /// does not matter. See [PostSortOption] constructor documentation for more
 /// details.
-double dayScore(PostFirestore post, GeoPoint userPosition) {
+double dayScore(PostFirestore post, GeoPoint _) {
```
gruvw, line 25: For unused parameters, does it work to simply prefix the argument name by an underscore instead of completely stripping out the name altogether ?

For example, here we could use `_userPosition` so it is still clear what the argument refers to.

If it works, could you apply this suggestion to all unused parameters, please ?

lib/viewmodels/posts_feed_view_model.dart
```diff
@@ -46,13 +46,9 @@ class PostsFeedViewModel extends AutoDisposeAsyncNotifier<List<PostDetails>> {
...
     );
 
-    // Wait for both futures to complete for optimisation
-    final List<dynamic> results = await Future.wait([
-      postsFirestoreFuture,
-      challengesFuture,
-    ]);
-    List<PostFirestore> postsFirestore = results[0];
-    final List<ChallengeFirestore> challenges = results[1];
+    List<PostFirestore> postsFirestore = await postsFirestoreFuture;
```
gruvw, line 49: This change decrease performance. The original author of this code wanted to have both future run at the same time asynchronously.

I think the solution for fixing the DCM hint here was to simply strip out the type and just have `final results = await Future.wait([` (if that works with type inference).

TheTexanCodeur: Fixed with [ac35e5f](https://github.com/ProximaEPFL/proxima/pull/262/commits/ac35e5ffef1b4166248b9ad89b4c8509b076cac5)

lib/services/sorting/post/post_score_functions.dart
TheTexanCodeur: When I try to apply your suggestion, I get a warning. Do we keep it like that then?

lib/services/sorting/post/post_score_functions.dart
TheTexanCodeur: Fixed with [3e3a507](https://github.com/ProximaEPFL/proxima/pull/262/commits/3e3a5072f110a02550d1168e151453441530cb40)

lib/services/database/upvote_repository_service.dart
TheTexanCodeur: I tried a new implementation with [9ebc786](https://github.com/ProximaEPFL/proxima/pull/262/commits/9ebc786a288ead702f4192a3fab954ea1326d700). What do you think about it?

lib/services/sorting/post/post_sorting_service.dart
TheTexanCodeur: I tried a new implementation with [9886fa4](https://github.com/ProximaEPFL/proxima/pull/262/commits/9886fa47af081891b783416283bf94f11ca43052). What's your opinion on it?

lib/services/database/upvote_repository_service.dart
gruvw: Yes, much better ty :+1:

lib/services/sorting/post/post_sorting_service.dart
gruvw: Much better, ty :+1:

lib/services/sorting/post/post_score_functions.dart
gruvw: No, then keep it the old way, I was wrong.

lib/views/components/async/loading_icon_button.dart
```diff
@@ -31,10 +31,9 @@ class LoadingIconButton extends HookWidget {
 
     final stillIconButton = IconButton(
       icon: Icon(icon, size: iconSize),
-      onPressed: () async {
+      onPressed: () {
         isLoading.value = LoadingState.pending;
-        await onClick();
-
+        onClick();
```
gruvw, line 36: This is not the same behavior as before. You can't just remove async and awaits. It won't produce the same output.

I advise you to read the following: https://dart.dev/codelabs/async-await

Otherwise, I discussed with @JoachimFavre and we might consider dropping the following DCM rule as it is not very constructive https://dcm.dev/docs/rules/common/avoid-passing-async-when-sync-expected/.

gruvw:
DCM rule `avoid-passing-async-when-sync-expected` dropped in 8408317.

I think we should revert the changes made against that rule. Sorry for the inconvenience caused. I did not think this rule was so intrusive and could lead to such weird code.

lib/views/components/async/loading_icon_button.dart
gruvw: See https://github.com/ProximaEPFL/proxima/pull/262#issuecomment-2106350762

TheTexanCodeur: I reverted the code back to how it was before with [4905711](https://github.com/ProximaEPFL/proxima/pull/262/commits/4905711405317b852f2c4d46faea227b410c3bab)

camillelnne (approved):
Thanks for the changes and the justification!
LGTM

lib/services/sorting/post/post_sorting_service.dart
```diff
@@ -25,32 +25,30 @@ class PostSortingService {
...
+        // a is considered smaller
+        return -1;
+      } else if (!putOnTopA && putOnTopB) {
+        return 1;
+      }
+
+      return defaultComparator(a, b);
     }
 
+    putOnTop.isEmpty
```
gruvw, line 48: Here what about the following:

```dart
final comparator = putOnTop.isEmpty ? defaultComparator : evaluateComparator;
```

lib/viewmodels/posts_feed_view_model.dart
```diff
@@ -46,13 +46,13 @@ class PostsFeedViewModel extends AutoDisposeAsyncNotifier<List<PostDetails>> {
...
 
-    // Wait for both futures to complete for optimisation
-    final List<dynamic> results = await Future.wait([
+    final results = await Future.wait([
       postsFirestoreFuture,
       challengesFuture,
     ]);
-    List<PostFirestore> postsFirestore = results[0];
-    final List<ChallengeFirestore> challenges = results[1];
+    List<PostFirestore> postsFirestore = results[0] as List<PostFirestore>;
```
gruvw, line 53: On those two lines the code doesn't need to specify the type of the variable twice.

You can just have `final varName = ... as Type;`

lib/views/pages/profile/profile_page.dart
```diff
@@ -42,8 +42,9 @@ class ProfilePage extends ConsumerWidget {
         ProfileInfoCard(
           content:
               "Here is a FAKE data comment on a super post that talks about something that is super cool and is located in a super spot that is very cosy and nice",
-          onDelete: () async {
```
gruvw, line 45: This change should be reverted. It was probably missed in 4905711.

TheTexanCodeur: If I revert this line I will get a warning of type "redundant 'async' keyword" by DCM. Should we really add the async keyword here even if we don't await anything for now?

lib/viewmodels/posts_feed_view_model.dart
TheTexanCodeur: I fixed one line with [e1d5922](https://github.com/ProximaEPFL/proxima/pull/262/commits/e1d5922b9b2629d9ee45df6a4fdfb8ee016550bf), but for the other one I don't think that I can remove the type specification as this variable can't be final (it's modified later in the code).

lib/services/sorting/post/post_sorting_service.dart
TheTexanCodeur: Thanks for the suggestion! Added with [f372091](https://github.com/ProximaEPFL/proxima/pull/262/commits/f3720915f1c939cf7ed82c8b81ce2625a74b8bdc)

lib/viewmodels/posts_feed_view_model.dart
gruvw: then you can use `var` instead of `final`

lib/views/pages/profile/profile_page.dart
gruvw: Do not only revert this line, revert the whole changes for this file.

Do you still get a warning ?

TheTexanCodeur: I reverted the file and and still get the warning. Should we leave the code like that?

gruvw: Yes, we will change this soon anyway as comment deletion is about to get implemented in UI.

gruvw (approved):

JoachimFavre (commented):
Those are good modifications overall. I just have some coments below, that should not take too much time fixing. Good job :)

lib/viewmodels/posts_feed_view_model.dart
```diff
@@ -46,13 +46,12 @@ class PostsFeedViewModel extends AutoDisposeAsyncNotifier<List<PostDetails>> {
       position,
     );
 
-    // Wait for both futures to complete for optimisation
```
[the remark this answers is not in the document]
TheTexanCodeur: Fixed with [eed255f](https://github.com/ProximaEPFL/proxima/pull/262/commits/eed255f9b05328add1b4172a188a909955b12464)

lib/services/sorting/post/post_sorting_service.dart
```diff
@@ -25,32 +25,27 @@ class PostSortingService {
...
-        if (putOnTopA && !putOnTopB) {
-          // a is considered smaller
-          return -1;
-        } else if (!putOnTopA && putOnTopB) {
-          return 1;
-        }
-
-        return defaultComparator(a, b);
-      };
+    int evaluateComparator(PostFirestore a, PostFirestore b) {
```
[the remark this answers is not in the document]
TheTexanCodeur: Fixed with [272343e](https://github.com/ProximaEPFL/proxima/pull/262/commits/272343e5bf67b9aa3c6c8e31ae3ab206c8cbe4af)

JoachimFavre (approved):
LGTM now! Good job! :)

Issue #256, opened 2024-05-10
Model to get personal comments


TheTexanCodeur:
Acceptance criteria:
- Have a repository that to store references to the comments that user has posted
- Have a method to get all the comments that a user have posted

PR #263, opened 2024-05-12
User comment model

yoannLafore:
Hello,
This PR aims to provide the model side that will allow the user to see the comments he has posted and delete them (issue #256).
Here the base idea behind the implementation. The user will have a collection `userComments` that stores references to the comments he has posted. Thus, when a comment is added on a post, the reference must also be added under the user document. Similarly when a comment is deleted.

More concretely, this PR does the following:

**Model classes for the user comments:**
This PR adds 3 classes to represent the reference to a comment the user has made: `UserCommentIdFirestore`, `UserCommentData`, `UserCommentFirestore`. 
In particular, `UserCommentData` is designed with the following attributes:
- `parentPostId`: The id of the post under which the comment has been posted
- `commentId`: The id of the comment that is referenced
- `content`: The content of the comment that is referenced. The content being present in the reference is an implementation choice. It allows to retrieve quickly the data to be displayed on the profile page. Without it, the alternative approach is to for each comment reference to query the full comment under the post, which adds reading cost and latency. Here, we only need to query the `userComments` collection to get the data to display in the profile page.

**Repository for the user comments:**
This PR adds the repository `UserCommentRepositoryService` that allows to manage the `userComments` collection. It contains methods to interact with the collection: `addUserComment`, `deleteUserComment`, `getUserComments`.

**Update the NewCommentViewModel:**
The `NewCommentViewModel` is updated to take account for this new implementation. Now, when a comment is added under a post, its reference is also added under the user document through the `UserCommentRepositoryService`.

**Mock generators:**  
This PR also adds two mock generators `UserCommentDataGenerator` and `UserCommentFirestoreGenerator` that are used to create mock instances of the model classes described above as well as providing utility functions to deal with the collection. Another small generator `FirestoreGenerator` is added with a function to provide a non-existing document snapshot which was necessary in a test.

**Tests for the model/repository:**
This PR adds tests for both the model classes and the repository service.
 

#### Acceptance criteria:
- [x] Have a repository that stores references to the comments that the user has posted
- [x] Have a method to get all the comments that a user have posted


lib/models/database/user_comment/user_comment_firestore.dart
```diff
@@ -0,0 +1,44 @@
...
+import "package:proxima/models/database/user_comment/user_comment_data.dart";
+import "package:proxima/models/database/user_comment/user_comment_id_firestore.dart";
+
+/// This class represents a comment from the user in firestore.
+class UserCommentFirestore {
+  static const String userCommentSubCollectionName = "userComments";
+
+  // The id of the firestore document that keeps the reference to the comment.
+  // Not the id of the comment itself.
+  final UserCommentIdFirestore id;
```
JoachimFavre, line 11: By humble curiosity, why did you choose to use a different ID as the `UserCommentIdFirestore`? Wouldn't it make sense to directly user `CommentIdFirestore` as id?
I feel like using a `CommentIdFirestore` as id would decrease the complexity by a non-negligible factor: you could remove it from `UserCommentData` and delete the class `UserCommentIdFirestore`.

lib/services/database/user_comment_repository_service.dart
```diff
@@ -0,0 +1,69 @@
...
+import "package:proxima/models/database/user/user_firestore.dart";
+import "package:proxima/models/database/user/user_id_firestore.dart";
+import "package:proxima/models/database/user_comment/user_comment_data.dart";
+import "package:proxima/models/database/user_comment/user_comment_firestore.dart";
+import "package:proxima/models/database/user_comment/user_comment_id_firestore.dart";
+import "package:proxima/services/database/firestore_service.dart";
+
+/// This class is responsible for managing the user's comments in the firestore database.
+/// This is the class to be used for displaying the user's comments on the profile page.
+class UserCommentReposittoryService {
```
JoachimFavre, line 12: There's a small typo here, Repository has a single t.
Tell me if you consider this is nitpicking, I have to admit I consider it to be important personally but I'm still learning not to be too nitpicky.

lib/viewmodels/new_comment_view_model.dart
```diff
@@ -82,7 +86,21 @@ class NewCommentViewModel
       voteScore: 0,
     );
 
-    await commentRepository.addComment(postId, commentData);
+    // Note: The addition of the comment unedr the post and the comment under the
```
JoachimFavre, line 89: Small typo:
```suggestion
    // Note: The addition of the comment under the post and the comment under the
```

lib/viewmodels/new_comment_view_model.dart
```diff
@@ -82,7 +86,21 @@ class NewCommentViewModel
...
+    // but the database is still consistent and won't crash the application.
+    final commentId = await commentRepository.addComment(postId, commentData);
+
+    final userCommentData = UserCommentData(
+      commentId: commentId,
+      parentPostId: postId,
+      content: content,
+    );
+
+    await userCommentRepository.addUserComment(currentUserId, userCommentData);
```
JoachimFavre, line 103: You add a reference here, but never delete it when the comment is deleted. In fact, I think that this is not the view-model's job to update this in the database, but directly the `CommentRepositoryService`'s. 
In other words, if I want to add another way to add comments to the code base, I don't want to have to duplicate this code; it should already be done by the model. If you follow this suggestion, make sure to adapt the documentation correctly: the `delete` function from the `UserCommentRepositoryService` should never be used, except by the delete `CommentRepositoryService` functions of. This is still a bit weird, maybe there is another way to encapsulate the delete function of `UserCommentRepositoryService` to make sure nobody can use it outside `CommentRepositoryService`. What do you think?

test/mocks/data/firestore_user_comment.dart
```diff
@@ -0,0 +1,48 @@
...
+import "user_comment_data.dart";
+
+class UserCommentFirestoreGenerator {
+  int _userCommentId = 0;
+  final UserCommentDataGenerator _userCommentDataGenerator;
+
+  UserCommentFirestoreGenerator({int seed = 0})
+      : _userCommentDataGenerator = UserCommentDataGenerator(seed: seed);
+
+  Future<List<UserCommentFirestore>> addComments(
```
JoachimFavre, line 17: I'm feeling like the two functions in this class have a very close meaning; the second one is almost `addComments(1, ..., ...)`. Why do you need the two? Couldn't they be merged?
What do you think?

test/mocks/data/user_comment_data.dart
```diff
@@ -0,0 +1,35 @@
...
+import "package:proxima/models/database/comment/comment_id_firestore.dart";
+import "package:proxima/models/database/post/post_id_firestore.dart";
+import "package:proxima/models/database/user_comment/user_comment_data.dart";
+
+class UserCommentDataGenerator {
+  final Random _random;
+
+  UserCommentDataGenerator({int seed = 0}) : _random = Random(seed);
+
+  List<UserCommentData> generateUserCommentData(int count) {
```
JoachimFavre, line 12: Same comment here.

test/services/database/user_comment_repository_test.dart
```diff
@@ -0,0 +1,174 @@
...
+          userCommentRepository,
+        );
+
+        final fetchedUserComments =
+            await userCommentRepository.getUserComments(user.uid);
+
+        expect(fetchedUserComments, unorderedEquals(userComments));
+      });
+
+      test("should throw an error if a document has missing fields", () async {
```
JoachimFavre, line 68: I just want to point your attention that you tested something very similar (slightly lower level) in `user_comment_data_test.dart`. Are you sure you want to keep this test? Tests are nice, but also complicate code maintanibility. 
I'll leave this to your judgement.

test/services/database/user_comment_repository_test.dart
```diff
@@ -0,0 +1,174 @@
...
+
+        final fetchedUserComments =
+            await userCommentRepository.getUserComments(user.uid);
+
+        expect(fetchedUserComments, [expectedUserComment]);
+      });
+    });
+
+    group("deleting user comments", () {
+      test("should delete a single user comment", () async {
```
JoachimFavre, line 114: Again, are you sure this test is really important? There's almost exactly the same one right after, line 138.

JoachimFavre (changes requested):
This is a great PR, which will be very useful. As usual with your code, even if there are many lines, it is quick and nice to read. 

I however still have some comments below, two of which are somewhat major. Moreover, I have the following two general suggestions:
1) I think it is interesting to reference the parameters in the documentation. So, writing  "This adds a post of id [postId] to the database" instead of "This adds a post to the database" for instance. This is not a big issue, and I don't think it is worth spending time looking for all occurrences to change in this PR. This is more a comment for the future.
2) I think you should add documentation to all the three classes you added to `test/mocks/data`.

lib/services/database/user_comment_repository_service.dart
yoannLafore: You are right !
I don't know how I missed it.
In my opinion, this is not nitpicky especially concerning class names that will be broadly used in the code :+1: 

Fixed in ac1e2f700737653f31dfcb66c7ae3f90df6ff1d8..

lib/viewmodels/new_comment_view_model.dart
yoannLafore: Thanks !

Fix in 643b7788b53fa6b6b1e97629276deb3208cbd70a.

test/services/database/user_comment_repository_test.dart
yoannLafore: You are right :+1: 
Looking back at it, I don't think this test adds anything to the robustness.

Thus I removed it in 064cd036cc03930d79298d0da7290c2494a1ef1a.

test/mocks/data/firestore_user_comment.dart
yoannLafore: I actually need the second one (`createMockUserComment`) when I have to get a `UserCommentFirestore` but doesn't need to have it in the database.

This happens in `user_comment_firestore_test.dart` for instance.
Merging the two and keeping only `addComments`, I would need to create a repository and add the comments to the database to get the mock data. And I don't think it makes a lot of sense for this test.

Thus, I personnaly prefer to keep the two functions here if that's ok for you ?

lib/models/database/user_comment/user_comment_firestore.dart
yoannLafore: At first, I thought it would make more sense to have them separated because all the comments are not in the same firestore collection.
But looking back at it, it adds quite a lot of complexity for not much thus I think that you are right.
Thanks for pointing it out !

Fix in 57a4b8d3caea5a2bbfa420938186a05afae385e2.

lib/viewmodels/new_comment_view_model.dart
yoannLafore: After thinking about it, I think that you are totally right !
However, this will require a substancial refactor.
Thus, I will remove the viewmodel update from this PR and perform the refactor in another one.
Thanks for catching important architecture mistake :+1: 

I remove this from the viewmodel in f39f4a5adc13daaf26eb67fb741f58a9cd716815.

yoannLafore: Note: Those function have slightly modified after removing the `UserCommentIdFirestore`.

I actually need the second one (`createMockUserComment`) when I have to get a `UserCommentFirestore` but doesn't need to have it in the database.

This happens in `user_comment_firestore_test.dart` for instance.
Merging the two and keeping only `addComments`, I would need to create a repository and add the comments to the database to get the mock data. And I don't think it makes a lot of sense for this test.

Thus, I personnaly prefer to keep the two functions here if that's ok for you ?

test/mocks/data/firestore_user_comment.dart
yoannLafore: Note: Those function have slightly modified after removing the `UserCommentIdFirestore`.

I actually need the second one (`createMockUserComment`) when I have to get a `UserCommentFirestore` but doesn't need to have it in the database.

This happens in `user_comment_firestore_test.dart` for instance.
Merging the two and keeping only `addComments`, I would need to create a repository and add the comments to the database to get the mock data. And I don't think it makes a lot of sense for this test.

Thus, I personnaly prefer to keep the two functions here if that's ok for you ?

test/services/database/user_comment_repository_test.dart
yoannLafore: I agree with you, this is testing to a bit lower level.
However, I designed that test to cover a part of the code that isn't covered otherwise.
So, do you think it is worth to keep it ?

JoachimFavre: I'm feeling like this is more of a test to add coverage than a real test bringing interest to the codebase. It maybe makes sense to get rid of the other test.
However, I know that you really like tests and coverage; and keeping both tests make sense. I let you do what you think is best.

test/mocks/data/firestore_user_comment.dart
JoachimFavre: What you say make sense. I like the fact that you use the second function explicitely in the first one now, to decrease code repetition. Good job :)

test/mocks/data/firestore_user_comment.dart
```diff
@@ -0,0 +1,53 @@
...
+    for (var i = 0; i < number; i++) {
+      final userComment = createMockUserComment();
+      userComments.add(
+        UserCommentFirestore(id: userComment.id, data: userComment.data),
+      );
+
+      await userCommentRepository.addUserComment(userId, userComments.last);
+    }
+
+    return userComments;
```
JoachimFavre, line 37: I feel like this implementation is quite inefficient: repetedly adding elements to a list is expensive, and waiting for the current comment to be added before moving onto the next one is not really efficient. I don't know if this is so important since this is a test, and maybe it is not a big bottleneck for test speed. I however suggest a differnet implementation, such as:
```suggestion
    final userComments = List.generate(number, (_) => createMockUserComment());
    await Future.wait(
      userComments.map(
        (comment) =>
            userCommentRepository.addUserComment(userId, userComments.last),
      ),
    );
    return userComments;
```

Note that I haven't tested this code, so I may have made a mistake.

test/mocks/data/user_comment_data.dart
JoachimFavre: I think what you did is completely fine, we have in multiple places two functions for mocks, one that generate a single element, other that generate a list of elements. I just think those two functions should have a more similar name. For instance, `generateUserCommentData` and `generateUserCommentDatas`.

test/services/database/user_comment_repository_test.dart
```diff
@@ -0,0 +1,138 @@
...
+
+        expect(fetchedUserComments, unorderedEquals(userComments));
+      });
+
+      test("should throw an error if a document has missing fields", () async {
+        final userComment = (await userCommentGenerator.addComments(
+          1,
+          user.uid,
+          userCommentRepository,
+        ))[0];
```
JoachimFavre, line 70: I think it would be nice to encapsulate this in a function `UserCommentGenerator::addComment`. This is done in other mocks data generator, and I feel like it brings a lot of readability to the code.
(By the way, if you decide to follow this suggestion, don't hesitate to use `.first` instead of `[0]` ;))

JoachimFavre (commented):
Code is much better now! Just some few fixes left.

lib/viewmodels/new_comment_view_model.dart
JoachimFavre: Nice, thank you for your work :)

test/mocks/data/user_comment_data.dart
yoannLafore: Oh ok I see.
That is a good idea indeed.
Done in 0b38809867c27818b007d46ce63717853e0f2bd4.

test/mocks/data/firestore_user_comment.dart
yoannLafore: As you said, I don't think that this is a very big deal given that this is for testing purposes.
But you implementation seems to be more concise thus I'll implement it :+1: 

Done in 143235bd5b1c3152d5158c343a883aa0e65216f7.

test/services/database/user_comment_repository_test.dart
yoannLafore: Good idea !

Done in 9a80cbc79307452f5ad323b5f58a601317d16d84.

test/services/database/user_comment_repository_test.dart
JoachimFavre: That's good for me, don't worry c:

JoachimFavre (approved):
LGTM, good job! :)

yoannLafore:
Thank you for all your amazingly great suggestion !

gruvw:
Is there a reason why this branch was not deleted ? Can we delete it now ?

yoannLafore:
You are right, I forgot to delete it.
I'll delete it now.

Issue #130, opened 2024-04-18
Fix release and apk build signature in CI


Aderfish:
The apks built from CI runs should always use the same signature so that it can be added once to firebase and the built apks always work.

Acceptance criteria:
- [ ] The apks built from CI work directly and connect to the database.

PR #268, opened 2024-05-14
Release ci fix

Aderfish:
Addresses #130,

This PR adds a keystore in a github secret so that the app is always signed with the same certificate on build. This means all apks built from now on will directly work with Firebase, without needing to manually add a certificate fingerprint in the settings every time.

This can be checked by running the following command on the build artifact.
```
keytool -printcert -jarfile *.apk 
```

Two repository secrets have been created:
`RELEASE_KEYSTORE_PROPERTIES` which contains the content of `key.properties`.
`B64_RELEASE_KEYSTORE` which contains the actual keystore. Since it is a binary file, it is stored in base 64 encoding so that it can easily be copied as text.

The android project has been modified to use a release signing key. The keystore properties are stored in `android/key.propertes`. During the build, this file is created from the secret. The keystore is imported from its secret and stored in `android/app/keystore/release.jks`.

You can find more information on the signing process [here](https://docs.flutter.dev/deployment/android#signing-the-app
).

`scripts/build` no longer works directly for local users. This can be fixed by manually creating the `android/key.properties` file locally. Its content could for example look like this to link to previously used debug keys.

```
storePassword=android
keyPassword=android
keyAlias=androiddebugkey
storeFile=/home/<user>/.android/debug.keystore
```

Or a developer could create a new certificate for release builds.

Acceptance criteria:
- [x] The apks built from CI work directly and connect to the database.


gruvw (approved):
Awesome work on this! Also thank you for taking the time to explain how it works to me IRL this morning.

PR #271, opened 2024-05-14
Profile color gradient

### LINKED ISSUE #255: Color of profile picture changes with points
Acceptance criteria:
- Define different colors linked to a certain amount of centauri points.
- The background of the user profile should change according to this color chart.
- This custom color should be visible in the post, comment and personal pages of the app

JoachimFavre:
Fixes #255.

This PR adds a gradient on the profile background, depending on the user Centauri points. Interpolation is done in the HSV (hue-saturation-value) space, to have something better linked to how human eye perceives colours. I let you take a look at the documentation to see how exactly interpolation is done. Note however that I use a different colour scheme depending on the application being in light or dark theme.
<p align="center">
  
  
</p

Importantly, I decided not to test [`DynamicUserAvatarViewModel:centauriToColor`](https://github.com/ProximaEPFL/proxima/pull/271/files#diff-e8c691d21522386573ab8af26268407964a3be3162dd87e88b9428f85accf814R38-R58), because it defines the UI. There is therefore no other way to test it than replicating its behaviour in the tests. However, this is bad if we want to change the UI colour scheme. Tell me if you think I should test it, and in what way.

Acceptance criteria:
- [x] Define different colors linked to a certain amount of centauri points.
- [x] The background of the user profile should change according to this color chart.
- [x] This custom color should be visible in the post, comment and personal pages of the app
- [x] Update the penpot accordingly


lib/models/ui/linear_segmented_hsv_colormap.dart
```diff
@@ -0,0 +1,84 @@
...
+  HSVColor call(int value) {
+    final indices = List.generate(stops.length, (index) => index);
+    final lowerBoundIdx = indices.lastWhereOrNull(
+      (index) => stops[index] <= value,
+    );
+    if (lowerBoundIdx == null) {
+      // value is less than the first stop
+      return colors.first;
+    }
+    if (lowerBoundIdx == stops.length - 1) {
```
TheTexanCodeur, line 67: In the comment you state that this case takes into account a value greater than or equal to the last stop but the if condition only checks equality. We could add an inequality check here or change the comment. What do you think?

lib/models/ui/linear_segmented_hsv_colormap.dart
```diff
@@ -0,0 +1,84 @@
...
+  /// [stops] is a list of values that define the stops of the colormap. Their corresponding
+  /// colors are interpolated between [hueStart] and [hueEnd] (in degrees), with a fixed
+  /// [saturation] and [value]. The distance between each stop colour is constant.
+  /// There must be at least two colour stops.
+  factory LinearSegmentedHSVColormap.uniform(
+    List<int> stops, {
+    double hueStart = 0,
+    double hueEnd = 330,
+    double saturation = 0.8,
+    double value = 0.5,
```
TheTexanCodeur, line 44: I don't think that this variables name is explicit enough. We could maybe rename it or be more precise in the comment above. What do you think?

JoachimFavre: I'm used to working with the HSV colour space, so I might be a bit biased indeed. However, I'm really unsure of how to name those variables in a more explicit manner. Hue, saturation and value are the name of the coordinates in this space (like red, green, blue or x, y, z). 
What other variable name would you consider?

TheTexanCodeur: Ok so let's keep it like that then!

PR #276, opened 2024-05-15
Refactor comment repositories

yoannLafore:
Hello,
This PR aims to provide a refactor of the comment repositories to abstract the comment referencing logic under the model layer (issue #277).

Concretely, this PR does the following:

**New  comment repository:**
The current `CommentRepositoryService` has been renamed to `PostCommentRepositoryService`.
Then, another repository named `CommentRepositoryService` is created. It manages the interactions between the post and user comment repositories. 
It exposes the following methods:
- `getPostComments`: Get all the `CommentFirestore` under a post
- `getUserComments`: Get all the `UserCommentFirestore` that a user has made
- `addComment`: Add comment under a post and add the corresponding reference to the user that posted the comment
- `deleteComment`: Delete a comment under a post and remove the comment reference in the user document
- `deleteAllComments`: Delete all the comments under a post and update the user references. To be consistent with the previous implementation of this method, this method takes as argument a `batch` that is used by the post repository to perform the deletion.

**Remove old providers:**
The old provider for the comment repository is replaced with the new one. And the provider for the user comments is simply removed. This is because only this new higher level repository should be used in the higher levels of the application (i.e. viewmodels).

**Tests:**
This PR adds tests for the new `CommentRepositoryService`. Those tests are intentionally lighter because the `PostCommentRepositoryService` and `UserCommentRepositoryService` are extensively tested in other files.
And this new `CommentRepositoryService` only abstracts those repositories with a little more logic added.

#### Acceptance criteria:
- [x] Have a global repository to abstract the post and user comments repositories
- [x] Addition/Deletion must also update the comment reference under the user
- [x] Have a method to delete all comments under a post and its references


JoachimFavre (commented):
Thank you so much, this refactor is great. 

However, I have to admit I don't really like the testing currently (you'll notice most of my comments are on this part). My biggest issue is that I don't think `comment_repository_test.dart`, `post_comment_repository_test.dart` and `user_comment_repository_test.dart` are meaningful here. I think those three files should be merged in a single test suite, that only test the `CommentRepositoryService`, not the two services it uses. This would allow multiple things:
1) This would allow to diminish a bit the 603 lines of tests we have concerning comment repositories.
2) This would allow to unify tests. We currently have very extensive tests in `post_comment_repository_test.dart`, and smaller ones in the two other files. Merging all the tests in a single test suite would allow to make sure we test the `deleteAll` function in regards of all comment-related collections at the same time and in the same manner. This would also simplify reasoning about tests.
3) This would improve separations of concern. In my opinion, current tests follow too closely the implementation, and make refactoring harder. As I mention in some of my other comments below, tests should not have to know `CommentRepositoryTest` uses two other repositories under the hood.

Also, continuing on the third point, I really think `PostCommentRepositoryService` and `UserCommentRepositoryService` should be made private. I know this is already a subject we considered, but together with my arguments here and the diff of `post_comment_repository_test.dart` which shows that we can still circumvent the fact that there is no provider for this repository, I think it is best for the code encapsulation.

lib/services/database/comment/comment_repository_service.dart
```diff
@@ -0,0 +1,115 @@
...
+import "package:proxima/models/database/user_comment/user_comment_data.dart";
+import "package:proxima/models/database/user_comment/user_comment_firestore.dart";
+import "package:proxima/services/database/comment/post_comment_repository_service.dart";
+import "package:proxima/services/database/comment/user_comment_repository_service.dart";
+import "package:proxima/services/database/firestore_service.dart";
+
+/// This service allows to interact with the comments of the application.
+/// It will allow to add, delete and retrieve comments from the firestore database.
+/// And it will handle the references between the comments and the users.
+class CommentRepositoryService {
```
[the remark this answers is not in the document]
yoannLafore: That's a good point indeed.
The two futures from `addComment` cannot be awaited in parallel because we need the result of `PostCommentRepository.addComment` to create the `UserCommentFirestore` to add afterward.

I now await the two futures in parallel in `deleteComment`, done in 5a70a13004e1964be15b78fb120a86772bd84382.

I believe the futures in `deleteAllComments` are already awaited in parallel but I have noticed that the code could be made more conscise. I do that in cef755a7eedb407045eb4de879b25d34c55d9567.

test/mocks/services/mock_comment_repository_service.dart
```diff
@@ -3,7 +3,9 @@ import "package:proxima/models/database/comment/comment_data.dart";
 import "package:proxima/models/database/comment/comment_firestore.dart";
 import "package:proxima/models/database/comment/comment_id_firestore.dart";
 import "package:proxima/models/database/post/post_id_firestore.dart";
-import "package:proxima/services/database/comment_repository_service.dart";
+import "package:proxima/models/database/user/user_id_firestore.dart";
+import "package:proxima/models/database/user_comment/user_comment_firestore.dart";
+import "package:proxima/services/database/comment/comment_repository_service.dart";
 
 class MockCommentRepositoryService extends Mock
```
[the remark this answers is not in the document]
yoannLafore: Good catch !
Fix in 27bfa4dee5f9924ddf2b98eb0112c16f38a39847.

test/services/database/comment/comment_repository_test.dart
```diff
@@ -0,0 +1,164 @@
...
+      // Check that it was deleted correctly
+      await checkPostAndUserCommentsEmpty();
+    });
+
+    test("should delete all the comments", () async {
+      // Add a comment
+      final commentData =
+          postCommentDataGenerator.createMockCommentData(ownerId: user.uid);
+
+      await commentRepo.addComment(post.id, commentData);
```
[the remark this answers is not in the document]
yoannLafore: You are right !
I have done a stronger test in 7127c3edd43576310aa1127f42dcc8afdeff4330.

test/services/database/comment/comment_repository_test.dart
```diff
@@ -0,0 +1,164 @@
...
+
+      final container = ProviderContainer(
+        overrides: [
+          firestoreProvider.overrideWithValue(fakeFirestore),
+        ],
+      );
+
+      commentRepo = container.read(commentRepositoryServiceProvider);
+
+      user = (await FirestoreUserGenerator.addUsers(fakeFirestore, 1)).first;
```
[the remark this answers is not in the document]
yoannLafore: Good idea !
Done in 4e5e467d6c92aa58cea5acd784df15cf9462de89.

test/services/database/comment_upvote_repository_test.dart
```diff
@@ -20,7 +20,7 @@ void main() {
   group("Testing comment upvote repository", () {
     late FakeFirebaseFirestore firestore;
     late PostRepositoryService postRepository;
-    late CommentRepositoryService commentRepository;
+    late PostCommentRepositoryService commentRepository;
```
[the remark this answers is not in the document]
yoannLafore: You are totally right !
Fix in 7c9888edf2eb1f7070ebd7ce0078a3807f8f2386.

test/services/database/comment/comment_repository_test.dart
```diff
@@ -0,0 +1,164 @@
...
+    test("should get the comments made by a user", () async {
+      final userComments =
+          await userCommentGenerator.addComments(3, user.uid, userCommentRepo);
+
+      final actualComments = await commentRepo.getUserComments(user.uid);
+
+      expect(actualComments, unorderedEquals(userComments));
+    });
+
+    test("should add a comment", () async {
```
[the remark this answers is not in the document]
JoachimFavre: Fix in ab9e43252f3ed9d74d59af212f110487f535ce20.

lib/services/database/comment/comment_repository_service.dart
```diff
@@ -0,0 +1,115 @@
...
+    UserIdFirestore userId,
+  ) =>
+      _userCommentRepo.getUserComments(userId);
+
+  // Note: The addition/deletion/allDeletion of the comment under the post and the comment under the
+  // the user document are not done fully atomically.
+  // This is because integrating atomicity would require a complex refactor
+  // that does not add much value to the application.
+  // In the worst case, the link between the comment and the user is lost,
+  // but the database is still consistent and won't crash the application.
```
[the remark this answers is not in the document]
JoachimFavre: Fixed in ddb68c45e3832711696137ed6eca502cb45a61a0.

lib/services/database/comment/comment_repository_service.dart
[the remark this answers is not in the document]
JoachimFavre: Good job :)

lib/services/database/comment/comment_repository_service.dart
```diff
@@ -0,0 +1,115 @@
...
+  ) async {
+    final comments = await getPostComments(parentPostId);
+
+    // The user comments are deleted in parallel
+    final userCommentsDeletion = comments.map((comment) async {
+      await _userCommentRepo.deleteUserComment(
+        comment.data.ownerId,
+        comment.id,
+      );
+    });
```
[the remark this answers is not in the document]
JoachimFavre: Yes indeed, this makes sense.

test/services/database/post_repository_test.dart
```diff
@@ -62,7 +62,8 @@ void main() {
       final dbPost = await postRef.get();
       expect(dbPost.exists, true);
 
-      final commentRepository = CommentRepositoryService(firestore: firestore);
+      final commentRepository =
+          PostCommentRepositoryService(firestore: firestore);
```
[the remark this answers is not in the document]
yoannLafore: As discussed with the group due to dart constraints, we chose to let those classes public but avoid using them outside of `CommentRepositoryService`.

I agree that we shouldn't use `PostCommentRepositoryService` here. This is fixed de4470d5f47019ab2600c040eb9f7d9c697ffee2.

test/mocks/data/firestore_comment.dart
```diff
@@ -16,7 +16,7 @@ class CommentFirestoreGenerator {
   Future<List<CommentFirestore>> addComments(
     int number,
     PostIdFirestore postId,
-    CommentRepositoryService commentRepository,
+    PostCommentRepositoryService commentRepository,
```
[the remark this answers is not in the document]
yoannLafore: Good idea !
This is done in de4470d5f47019ab2600c040eb9f7d9c697ffee2.

test/services/database/comment/comment_repository_test.dart
```diff
@@ -0,0 +1,316 @@
...
+
+      // Check that the user comments have been added
+      for (user in users) {
+        final userComments = await commentRepo.getUserComments(user.uid);
+
+        expect(userComments, isNotEmpty);
+      }
+
+      return (users, postComments);
+    }
```
JoachimFavre, line 240: This seems close in behaviour with `CommentFirestoreGenerator:addCommentsForUsers`. Is there a way to diminish code repetition?

test/services/database/comment/comment_repository_test.dart
```diff
@@ -0,0 +1,316 @@
...
+
+    group("deleting comments", () {
+      test("should delete a comment", () async {
+        // Add a comment
+        final commentData =
+            postCommentDataGenerator.createMockCommentData(ownerId: user.uid);
+
+        final commentId = await commentRepo.addComment(post.id, commentData);
```
JoachimFavre, line 248: This can be simplified using `CommentFirestoreGenerator:addComment`.

yoannLafore:
Thanks for all your great suggestions :tada: 
I followed your idea to merge the tests.
This as indeed a good think and it makes the testing codebase more flexible for the future.
There is just a test from `user_comment_repository_test.dart` that was removed as it was not very useful and doesn't incorporate well. This test was "should throw an error if a document has missing fields".

test/services/database/comment/comment_repository_test.dart
```diff
@@ -0,0 +1,316 @@
...
+
+        // Check that it was added correctly
+        await checkPostAndUserCommentsNotEmpty();
+
+        // Delete the comment
+        await commentRepo.deleteComment(post.id, commentId, user.uid);
+
+        // Check that it was deleted correctly
+        await checkPostAndUserCommentsEmpty();
+      });
```
JoachimFavre, line 258: I wonder if this test is really useful, considering the following test does the same but is more robust.

test/services/database/comment/comment_repository_test.dart
```diff
@@ -0,0 +1,316 @@
...
+
+      test("should delete a comment when they are multiple comments", () async {
```
JoachimFavre, line 260: Small typo:
```suggestion
      test("should delete a comment when there are multiple comments", () async {
```

test/services/database/comment/comment_repository_test.dart
```diff
@@ -0,0 +1,316 @@
...
+
+        // Check that all the comments were deleted
+        final actualPostComments = await commentRepo.getPostComments(post.id);
+        expect(actualPostComments, isEmpty);
+
+        for (user in users) {
+          final userComments = await commentRepo.getUserComments(user.uid);
+          expect(userComments, isEmpty);
+        }
+      });
```
JoachimFavre, line 313: I think it is interesting to add a comment that is not deleted by the `deleteAll`:
```suggestion
      test("should delete all the comments", () async {
        const nbUsers = 5;
        late List<UserFirestore> users;

        // All users have a comment under the post
        (users, _) = await addCommentsForUsers(nbUsers);
        // The first user has a post in the other post
        final (_, expectedFirstUserComment) = await commentGenerator.addComment(
          otherPost.id,
          users.first.uid,
          commentRepo,
        );

        final firstUserComments = await commentRepo.getUserComments(
          users.first.uid,
        );
        expect(firstUserComments.length, equals(2));

        final batch = fakeFirestore.batch();
        await commentRepo.deleteAllComments(post.id, batch);
        await batch.commit();

        // Check that all the comments were deleted
        final actualPostComments = await commentRepo.getPostComments(post.id);
        expect(actualPostComments, isEmpty);

        // The first user is the only one with a comment left
        final actualFirstUserComments = await commentRepo.getUserComments(
          users.first.uid,
        );
        expect(actualFirstUserComments, equals([expectedFirstUserComment]));

        for (user in users.skip(1)) {
          final userComments = await commentRepo.getUserComments(user.uid);
          expect(userComments, isEmpty);
        }
      });
```

test/services/database/comment_repository_test.dart
```diff
@@ -1,298 +0,0 @@
...
-
-        // Check that the comment count was updated correctly
-        final postDoc = await postDocument.get();
-        final post = PostFirestore.fromDb(postDoc);
-
-        expect(post.data.commentCount, equals(alreadyPresentCommentsCount - 1));
-      });
-
-      test(
-          "should thrown an error and do nothing if the comment does not exist",
```
JoachimFavre, line 254: This test is missing in the new file. Is it expected?

test/services/database/comment/comment_repository_test.dart
yoannLafore: You are right !
I reduce code dupllication in b964bf49e35e46c295a9202a00d36c86f67fc912.

test/services/database/comment/comment_repository_test.dart
yoannLafore: Good catch !
Done in 4758cd2e405d25f4aea9fb700faaf046ec6b57d6.

test/services/database/comment_repository_test.dart
JoachimFavre: Fixed in ba41cb876b1e7ac78e33e5302cdf61a525f319d3.

test/services/database/comment/comment_repository_test.dart
yoannLafore: Looking again at it, I don't think that it is necessary anymore.
Removed in 4e7d49eecd82658bef0ebbab76c9cbb3bacdd8a1.

test/services/database/comment/comment_repository_test.dart
yoannLafore: Good idea !
Done in 310ffc73223c2f15a0af61690b660d8d6afe80df.

test/services/database/comment_repository_test.dart
yoannLafore: Good initiative !
Your implementation looks good to me.
Thanks :+1:

JoachimFavre (approved):
This was quite a long journey, a struggle ending at 1am, but we made it! Good job :D

LGTM! 🐘

yoannLafore:
Thanks for your great help !

gruvw:
Is there a reason why this branch was not deleted ? Can we delete it now ?

PR #314, opened 2024-05-17
Offline Circular Value Adaptation

### LINKED ISSUE #278: Add timeout on circular values for offline mode
Acceptance criterias:
- [ ] Have a timeout on the circular value
- [ ] Display a snackbar message on reaching timeout
- [ ] Have a retry button instead of spinning forever
- [ ] Adapt main circular value uses to have the new timeout

gruvw:
Fixes #278 

### Offline Circular Value Adaptation

This PR aims to improve the circular value by adding a default timeout after which a pop-up error is displayed instead of loading indefinitely, notably when offline.

This required quite a lot of adaptation of the `CircularValue` helper widget and its use around the app.
It is due to the fact that the old `CircularValue` took an `AsyncData` instead of a future as an argument, and this wrapper type did not support timeouts.
Therefore, I created a new `Future` wrapper type called `FutureRes` (supporting timeouts) and we can wrap any `Future` using the simple `.mapRes()` extension provided in `lib/views/helpers/types.dart`.

#### Acceptance criteria

- [X] New system using future wrapping and new type `FutureRes`
- [X] Full refactor of the `CircularValue` to accept a mapped future in `FutureRes` instead of `AsyncData`
- [X] Offline timeout on circular value
- [X] Optional retry component when offline timeout on circular value (use the fallback builder)
- [X] Check that new circular value works everywhere as expected
- [X] Write tests for new circular value using timeout

#### Result

The resulting offline error dialog:

#### Public Service Announcement

The new way to use a provider's result using the circular value (to display the loading indicator) is the following:

```dart
final asyncPosts = ref.watch(postsFeedViewModelProvider.future).mapRes();
...
CircularValue(
    future: asyncPosts,
    ...
```

This circular value will timeout after `CircularValue.offlineTimeout` seconds (8 seconds) and display the optional fallback builder instead of the loading indicator if the timeout (or an error) occurs. Check the `fallbackBuilder` field of the `CircularValue`.
If you want a basic error handling with a retry button, you can check out the `ErrorRefreshPage` that was added in #334, and set this as the `fallbackBuilder` (full example for the post feed in `lib/views/pages/home/content/feed/post_feed.dart`).

**Note**: this PR also bumps the project version to `v0.4.2` (Second sprint of Milestone 4).

**Note**: this PR also fixes the flutter android warning "OnBackInvokedCallback is not enabled for the application.", check `AndroidManifest.xml` or https://stackoverflow.com/questions/76487677/onbackinvokedcallback-is-not-enabled-for-the-application-in-set-androidenableo for more details.

Please don't hesitate to ask me questions if something is not clear :)


test/mocks/overrides/override_posts_feed_view_model.dart
```diff
@@ -39,8 +40,10 @@ final mockNonEmptyHomeViewModelOverride = [
 final mockLoadingHomeViewModelOverride = [
   postsFeedViewModelProvider.overrideWith(
     () => MockPostsFeedViewModel(
-      // Future.any([]) will never complete and simulate a loading state
-      build: () => Future.any([]),
+      build: () async {
+        await Future.delayed(Durations.long1);
+        return [];
+      },
```
CHOOSEIT, line 46: I am curious, why do you make it able to complete after 450 ms ?

Previously we had a never ending future

test/mocks/providers/provider_circular_value.dart
```diff
@@ -1,15 +1,19 @@
...
     child: MaterialApp(
       home: Scaffold(
         body: CircularValue(
-          value: value,
+          future: future().mapRes(),
           builder: (context, data) => const Text("Completed"),
-          fallbackBuilder: (context, error) => const Text("Strange Error"),
+          fallbackBuilder: fallbackBuilder ?? CircularValue.defaultFallback,
         ),
       ),
```
CHOOSEIT, line 18: Small suggestion, since you are adding the `fallbackBuilder` shouldn't we also add the `builder` for completeness ?

test/mocks/overrides/override_posts_feed_view_model.dart
CHOOSEIT: This is also linked to the Flutter timers that you had to clear in `Display circular value on loading` (`test/views/pages/home/static_home_test.dart`).

With the never ending future, you are not forced to clear the flutter timers 😄

CHOOSEIT (changes requested):
Really great work 🎉 

I really like your class `FutureRes` and your extension function on Future. This is really well thought 💯 

Thank you 😄

test/mocks/providers/provider_circular_value.dart
gruvw: Sure, not sure why we should do it now if nobody needs it at the moment, but I added it in for completeness. 893a9d0c

I'm more in favor of adding those things when they are needed rather than upfront, because more often than not, nobody will need them.

test/mocks/overrides/override_posts_feed_view_model.dart
gruvw: To stay consistent with other tests, I now use `Durations.short1` (see b81aff5e) to avoid confusion about the exact delay of the future. We just want a short wait to try the loading.

I don't see how it is possible to test this without triggering an error if we don't clear the flutter timers.
If you manage to do it without changing the testing behavior, I would be happy if you can commit your solution on this branch.

lib/views/components/async/circular_value.dart
```diff
@@ -19,43 +22,96 @@ class CircularValue<T> extends StatelessWidget {
...
+  static const offlineTimeout = Duration(seconds: 8);
+
+  static Widget defaultFallback(BuildContext _, Object __) =>
+      const SizedBox.shrink();
+
   /// Constructor for the [CircularValue] widget.
   /// [value] is the underlying [AsyncValue] that controls the display.
   /// [builder] is the widget to display when the [value] is [AsyncValue.data].
   /// [fallbackBuilder] is the widget to display when the [value] is [AsyncValue.error].
   /// The default [fallbackBuilder] is an empty [SizedBox].
```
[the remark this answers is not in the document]
gruvw: Absolutely, totally missed that. Done in 49f23661

test/mocks/overrides/override_posts_feed_view_model.dart
CHOOSEIT: I see.

> I don't see how it is possible to test this without triggering an error if we don't clear the flutter timers. If you manage to do it without changing the testing behavior, I would be happy if you can commit your solution on this branch.

With the Future.wait([]), the test pass because we don't create a future, so none are pending when ending the test.

But if we keep the `Durations.short1`, yes we need to clear the flutter timers 👍🏼 

Thank you

gruvw: `Future.wait([])` is not a never ending future, right ? It will complete the next cycle.

gruvw: I think you confused `Future.wait([])` with `Future.any([])`.

Try the following in https://dartpad.dev:

```dart
void main() async {
  print("Starting.");
  await Future.wait([]); // completes instantly
  print("Done waiting on []");
  await Future.any([]); // never ending future
  print("This will not get logged.");
}
```

Aderfish (commented):
Good job on this PR, this is a very important functionality and I see that it was not easy to add. I have just left one comment on naming but otherwise the code looks good to me.

lib/views/helpers/types.dart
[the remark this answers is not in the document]
gruvw: Yes, I absolutely agree with you.

I named it `Result` to stay consistent with what most functional programming languages offer.

Refactored in f7a504a1.

Aderfish (approved):

CHOOSEIT (approved):
Looks good to me 💯

Issue #299, opened 2024-05-17
Ranking view model


CHOOSEIT:
Acceptance criterias:
- [x] Retrieve users that have most points from DB
- [x] Create new UI model class to expose user ranking fields
- [x] Write a viewmodel with public API to get the top 50 users of the app
- [x] Write another function to get the current user (potentially with its own UI model class)

gruvw:
- Retrieve users that have most points from DB
- Create new UI model class to expose user leaderboard fields
- Write a viewmodel with public API to get the top 50 users of the app
- Write another function to get the current user (potentially with its own UI model class)

Issue #298, opened 2024-05-17
Ranking UI


CHOOSEIT:
Acceptance Criteria:
- [ ] Create the UI foundation to display the leaderboard.
- [ ] Update the Penpot design accordingly.

PR #319, opened 2024-05-18
Ranking UI foundation

CHOOSEIT:
This PR aims to lays the foundations for the tasks #298 and #299.

It provides and test two data classes `RankingElementDetails` and `RankingDetails` that contains the necessary information needed to display the ranking page.

`RankingElementDetails` provide the necessary information to an entry in the ranking page:
image

`RankingDetails` contains a list of `RankingElementDetails` along with the `RankingElementDetails` information about the current user.
image

Theses images comes from our mockup

### DIFF
--- a/lib/models/ui/ranking/ranking_details.dart
+++ b/lib/models/ui/ranking/ranking_details.dart
@@ -2,6 +2,11 @@ import "package:flutter/foundation.dart";
 import "package:proxima/models/ui/ranking/ranking_element_details.dart";
 
 /// A class that stores data for the ranking widget.
+/// Elements in [rankElementDetailsList] should have non-null [userRank].
+///
+/// Note that we don't ensure that the [userRank] are not empty as we want to leave flexibility
+/// to the developer. Moreover, when used in the UI, this will yield a runtime error
+/// if the [userRank] is null.
 @immutable
 class RankingDetails {
   const RankingDetails({
--- a/lib/models/ui/ranking/ranking_details.dart
+++ b/lib/models/ui/ranking/ranking_details.dart
@@ -1,0 +1,32 @@
+import "package:flutter/foundation.dart";
+import "package:proxima/models/ui/ranking/ranking_element_details.dart";
+
+/// A class that stores data for the ranking widget.
+@immutable
+class RankingDetails {
+  const RankingDetails({
+    required this.userRankElementDetails,
+    required this.rankElementDetailsList,
+  });
+
+  /// Details of the user's rank.
+  final RankingElementDetails userRankElementDetails;
+
+  /// List containing details the ranking elements.
+  final List<RankingElementDetails> rankElementDetailsList;
+
+  @override
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
+
+    return other is RankingDetails &&
+        other.userRankElementDetails == userRankElementDetails &&
+        other.rankElementDetailsList == rankElementDetailsList;
+  }
+
+  @override
+  int get hashCode => Object.hash(
+        userRankElementDetails,
+        rankElementDetailsList,
+      );
+}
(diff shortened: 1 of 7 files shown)


lib/models/ui/ranking/ranking_details.dart
```diff
@@ -0,0 +1,37 @@
...
+/// to the developer. Moreover, when used in the UI, this will yield a runtime error
+/// if the [userRank] is null.
+@immutable
+class RankingDetails {
+  const RankingDetails({
+    required this.userRankElementDetails,
+    required this.rankElementDetailsList,
+  });
+
+  /// Details of the user's rank.
```
gruvw, line 17: Can you add here (in the documentation) the fact that the `userRank` attribute of this `RankingElementDetails` will be null.

lib/models/ui/ranking/ranking_details.dart
```diff
@@ -0,0 +1,37 @@
...
+  final RankingElementDetails userRankElementDetails;
+
+  /// List containing details the ranking elements.
+  final List<RankingElementDetails> rankElementDetailsList;
```
gruvw, line 21: Can you add here (in the documentation) the fact that the `userRank` attribute of every `RankingElementDetails` in this list wont be null.

lib/models/ui/ranking/ranking_element_details.dart
```diff
@@ -0,0 +1,44 @@
...
+/// The [userRank] is nullable to allow the current user to not have a rank.
+@immutable
+class RankingElementDetails {
+  const RankingElementDetails({
+    required this.userDisplayName,
+    required this.userUserName,
+    required this.centauriPoints,
+    required this.userRank,
+  });
+
```
gruvw, line 13: I think here we should have the user id too. We might use it for the avatar and the profile pop up.

gruvw (changes requested):

gruvw:
Great start on this PR, a just noticed a few things to change.

lib/models/ui/ranking/ranking_details.dart
CHOOSEIT: You are completely right !

Added in: [8a55765](https://github.com/ProximaEPFL/proxima/pull/319/commits/8a55765d5958a96c8109fae1ca8da60b472f64cf)

Thank you !

lib/models/ui/ranking/ranking_details.dart
CHOOSEIT: You are completely right !

Added in: [8a55765](https://github.com/ProximaEPFL/proxima/pull/319/commits/8a55765d5958a96c8109fae1ca8da60b472f64cf)

**Note:** that I also added assertion test on the constructor to ensure that the `userRank` are non-null, because I realised that in the UI, I wont test it and it may not yield an error.

Thank you !

lib/models/ui/ranking/ranking_element_details.dart
CHOOSEIT: I've made some local progress on the user interface and it seems that the for the moment the `userId` is not necessary and I am able to display the ranking card information along with the pop up.

I will follow up on this discussion, if I have any update :)

Thank you :)

lib/models/ui/ranking/ranking_details.dart
gruvw: Asserts are indeed a great addition, thank you :)

lib/models/ui/ranking/ranking_element_details.dart
gruvw: Oh right. I just double-checked, I thought that `UserProfilePopUp` was taking a user id as parameter but apparently not :+1:

gruvw (approved):
Awesome start :) We can now both work on our task #298 and #299 on our own branch from this common base.

We'll merge this when our 2 other PRs are ready to merge.

PR #320, opened 2024-05-18
My posts option on the map

### LINKED ISSUE #291: My posts on the map

JoachimFavre:
Fixes #291.

This PR make the chips above the map work. It is now possible to click on them to select an option. The default behaviour of getting the `Nearby` posts still work, but the chip giving `My posts` now also works. Note that the pins are refreshed every time a new post is created, and one is deleted. 

This is tested by clicking on the chips and verifying the correct pins are output by the view-model. Since we also have a test that verifies the pins displayed on the map correspond to the ones of the view-model, I felt this was the best and most intuitive way to do. Tell me if you wish me to instead verify the pins on the map directly.

Finally, I had to add a `PostGenerator:createUserPosts` function to abstract away data creation from my testing. 

Have fun reading this PR, I look forward to your comments! :)

Acceptance criteria:
- [x] Display the posts of the user on the map
- [x] Enable the choice chip to choose between the available options and the newly added option


test/mocks/providers/provider_homepage.dart
```diff
@@ -44,10 +44,12 @@ final loadingHomePageProvider = ProviderScope(
...
 ProviderScope homePageFakeFirestoreProvider(
   FakeFirebaseFirestore firestore,
-  MockGeolocationService geoLocationService,
-) {
+  MockGeolocationService geoLocationService, {
+  List<Override> additionalOverrides = const [],
+}) {
   return ProviderScope(
     overrides: [
+      ...additionalOverrides,
```
CHOOSEIT, line 52: I think this is a good idea to add a `additionalOverrides`, but at the same time, I think it breaks the idea that I have of the mocks providers that we have (idea of ready to use). 

I understand that it adds some flexibility to our mocks providers. However, to keep our mock provider file consistent, shouldn't we duplicate this part and create a provider that directly adds the override that you use in your tests?

As for the moment, your additional override has a single usage, but if we want to expand your tests or use the same mock provider, we may have to duplicate the `additionalOverrides`.

I understand both sides: flexibility / duplication of providers.

I am curious to hear your thoughts about it 😄

test/views/pages/map/dynamic_map_test.dart
```diff
@@ -0,0 +1,286 @@
...
+        equals(MapSelectionOptionsViewModel.defaultMapOption),
+      );
+    });
+
+    /// Verifies that the [option] is indeed the current option of the view-model
+    /// (in the view of the [container]), and that the pins returned by the view-model
+    /// are indeed the one expected in [expectedPostsForOption]. If [expectedPostForOption]
+    /// does not contain the [option], then it is assumed that the expected posts
+    /// are an empty list.
+    Future<void> testOption(
```
CHOOSEIT, line 146: I think we could use a more explicit name for this function. Maybe something like: `testMapSelectionOptionAndPins`

test/views/pages/map/dynamic_map_test.dart
```diff
@@ -0,0 +1,286 @@
...
+        await tester.pumpAndSettle();
+      },
+    );
+
+    testNavigation(
+      testTitle: "Pins refresh after post deletion",
+      optionToTest: MapSelectionOptions.myPosts,
+      expectedPinDelta: -1,
+      protocol: (tester, container) async {
+        final profileButton = find.byType(UserAvatar);
```
CHOOSEIT, line 268: I think that we could use the key that is present on the user avatar.

```suggestion
        final profileButton = find.byKey(HomeTopBar.profilePictureKey);
```

test/views/pages/map/dynamic_map_test.dart
```diff
@@ -0,0 +1,286 @@
...
+        expect(pinsEnd, hasLength(expectedPins.length + expectedPinDelta));
+      });
+    }
+
+    testNavigation(
+      testTitle: "Pins refresh after post creation",
+      optionToTest: MapSelectionOptions.nearby,
+      expectedPinDelta: 1,
+      protocol: (tester, container) async {
+        await tester.tap(find.text("New post"));
```
CHOOSEIT, line 241: I think that this line can be removed.

```suggestion
```

CHOOSEIT (changes requested):
Really clean PR 💪🏼 , I don't have much to say. Great job 💯 .
This PR really gives our map another dimension. 💪🏼 

This was really a good idea to abstract the option selection view model 👍🏼 

And your tests were really pleasant to read 💯  Great job !

Thank you !

_Note: I saw that the original authors of some classes in which you worked on haven't added  any class documentation. Could you add a little one to them, please ?
I am thinking: `NewPostViewModel` and `MapSelectionOptionChips`._

test/mocks/providers/provider_homepage.dart
JoachimFavre: This is a very good point, I did not consider it at all. I completely agree with you, I'm just afraid it may decrease slightly the readability of the code. Therefore, please tell me what you think of the variable names I used in fe04e87d9d49d38286c9dcd78357133a4fc53a6d.

test/views/pages/map/dynamic_map_test.dart
JoachimFavre: Oh, this is an excellent catch good job! This line should definitely not have found its way here, thank you so much for having found it!
Fixed in e9e5f52e206a6135ce9298589efdf5a05c8c5d49.

test/views/pages/map/dynamic_map_test.dart
JoachimFavre: Fixed in 5f038357a46beb16c994a21b3577faaf0e59d7a0.

JoachimFavre:
> _Note: I saw that the original authors of some classes in which you worked on haven't added  any class documentation. Could you add a little one to them, please ?_
> _I am thinking: `NewPostViewModel` and `MapSelectionOptionChips`._ 

I'm not sure this was really the job of my PR to add documentation to `NewPostViewModel`. This was however very quick to do and will probably not create any merge conflict, so I guess I can do this indeed. Fixed in 7ee7f51865c6cab370c71a81480d81c6702df214 and 06292660a1e420ad362a05073ae6caebce785161.

test/views/pages/map/dynamic_map_test.dart
CHOOSEIT: That's perfect ! Thank you ! 👍🏼

CHOOSEIT:
> I'm not sure this was really the job of my PR to add documentation to `NewPostViewModel`. This was however very quick to do and will probably not create any merge conflict, so I guess I can do this indeed. Fixed in [7ee7f51](https://github.com/ProximaEPFL/proxima/commit/7ee7f51865c6cab370c71a81480d81c6702df214) and [0629266](https://github.com/ProximaEPFL/proxima/commit/06292660a1e420ad362a05073ae6caebce785161).

Yeah, I agree. However, when we can improve the documentation a little, I think it could be a good idea to do so instead of doing a separate PR for the entire code base. 😄 👍🏼 

Thank you ! 💯 😄

CHOOSEIT (approved):
Looks good to me now !

Great work ! 💯 

Thank you !

PR #321, opened 2024-05-18
Challenges option on the map

### LINKED ISSUE #296: My challenges on the map
Acceptance criteria:
 
- [ ] Display the challenges of the user on the map
- [ ] Enable the choice chip to choose between the available options and the newly added option

JoachimFavre:
Fixes #296.

This PR adds the functionality of the "Challenges" selection chip above the map. Only uncompleted challenges are displayed on the map. Once a challenge is completed, the pins are refreshed. This is also completely tested.

This PR strongly depends on the code I added in #320. I already had this PR in mind, so I made some very modular code and tests immediately, hence this PR size. 

Happy PR reviewing :)

Acceptance criteria:
- [x] Display the challenges of the user on the map
- [x] Enable the choice chip to choose between the available options and the newly added option

### DIFF
--- a/lib/viewmodels/map/map_pin_view_model.dart
+++ b/lib/viewmodels/map/map_pin_view_model.dart
@@ -2,6 +2,7 @@ import "package:google_maps_flutter/google_maps_flutter.dart";
 import "package:hooks_riverpod/hooks_riverpod.dart";
 import "package:proxima/models/database/post/post_firestore.dart";
 import "package:proxima/models/ui/map_pin_details.dart";
+import "package:proxima/services/database/challenge_repository_service.dart";
 import "package:proxima/services/database/post_repository_service.dart";
 import "package:proxima/services/sensors/geolocation_service.dart";
 import "package:proxima/viewmodels/login_view_model.dart";
@@ -21,6 +22,8 @@ class MapPinViewModel extends AsyncNotifier<List<MapPinDetails>> {
         return _getNearbyPosts();
       case MapSelectionOptions.myPosts:
         return _getUserPosts();
+      case MapSelectionOptions.challenges:
+        return _getUserChallenges();
       default:
         return List.empty();
     }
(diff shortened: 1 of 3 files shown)


lib/viewmodels/map/map_pin_view_model.dart
```diff
@@ -61,6 +65,35 @@ class MapPinViewModel extends AutoDisposeAsyncNotifier<List<MapPinDetails>> {
...
+    final postRepository = ref.watch(postRepositoryServiceProvider);
+    final challengeRepostory = ref.watch(challengeRepositoryServiceProvider);
+    final userId = ref.watch(validLoggedInUserIdProvider);
+    // Only doing a read here, to decrease the number of database reads
+    // (we don't want to re-read the challenges when the position changes).
+    final position = await ref.read(livePositionStreamProvider.future);
+
+    if (position == null) {
+      return List.empty();
+    }
```
CHOOSEIT, line 79: This is a great idea to only read on the position.

Moreover if we just want the read the position we can use the `geolocationServiceProvider` provider that directly gives us a non-null `GeoPoint`.

```suggestion
    final position =
        await ref.read(geolocationServiceProvider).getCurrentPosition();
```

test/views/pages/map/dynamic_map_test.dart
```diff
@@ -120,9 +125,28 @@ void main() {
       );
       await setPostsFirestore(userPosts, fakeFirestore);
 
+      challenges = [nearbyPosts.first, farPosts.first, userPosts.first];
```
CHOOSEIT, line 128: I think that the `userPosts.first` post cannot be chosen as challenge in theory as it is one of the current user post. We may want to change it by another 😄

CHOOSEIT (changes requested):
Clean. 

This PR really leverage the strong foundation that you built in #320. 

Great work ! 💯 

Thank you

lib/viewmodels/map/map_pin_view_model.dart
JoachimFavre: Good catch, I didn't think of that! Great job :)
Fixed in e26a8ca23ae8c739445548da158b4fd4a58c7463.

CHOOSEIT (approved):
Looks good to me now !

Great work 💯  Thank you !

PR #322, opened 2024-05-19
Ranking UI

CHOOSEIT:
Fixes: #298 
This PR aims to provide the UI for the ranking page.

image

**Note:**  We are working with Lucas in parallel on the view and the view model. So we are trying to reduce the amount of conflicts on our tasks. In this PR, all the mock datas are in the view for the moment.

Acceptance Criteria:

 - [x] Create the UI foundation to display the leaderboard.
 - [x] Update the Penpot design accordingly.

### DIFF
--- a/lib/views/pages/home/content/ranking/ranking_card.dart
+++ b/lib/views/pages/home/content/ranking/ranking_card.dart
@@ -35,19 +35,19 @@ class RankingCard extends StatelessWidget {
 
   final RankingElementDetails rankingElementDetails;
 
-  void onTapUserPopUp(BuildContext context) {
-    showDialog(
-      context: context,
-      builder: (BuildContext context) => UserProfilePopUp(
-        displayName: rankingElementDetails.userDisplayName,
-        username: rankingElementDetails.userUserName,
-        centauriPoints: rankingElementDetails.centauriPoints,
-      ),
-    );
-  }
-
   @override
   Widget build(BuildContext context) {
+    void onTapUserPopUp() {
+      showDialog(
+        context: context,
+        builder: (BuildContext context) => UserProfilePopUp(
+          displayName: rankingElementDetails.userDisplayName,
+          username: rankingElementDetails.userUserName,
+          centauriPoints: rankingElementDetails.centauriPoints,
+        ),
+      );
+    }
+
     final cardColor = switch (rankingElementDetails.userRank) {
       1 => _rankOneColor,
       2 => _rankSecondColor,
@@ -117,7 +117,7 @@ class RankingCard extends StatelessWidget {
       clipBehavior: Clip.hardEdge,
       color: cardColor,
       child: InkWell(
-        onTap: () => onTapUserPopUp(context),
+        onTap: () => onTapUserPopUp(),
         child: Row(
           children: [
             userRankText,
--- a/lib/views/pages/home/content/ranking/ranking_card.dart
+++ b/lib/views/pages/home/content/ranking/ranking_card.dart
@@ -9,13 +9,6 @@ import "package:proxima/views/components/content/user_avatar/user_avatar.dart";
 /// The card is clickable and opens a user profile pop-up.
 /// The card's color can be customized.
 class RankingCard extends StatelessWidget {
-  const RankingCard({
-    super.key,
-    required this.rankingElementDetails,
-  });
-
-  final RankingElementDetails rankingElementDetails;
-
   static const userRankrankTextKey = Key("userRankrankText");
   static const userRankDisplayNameTextKey = Key("userRankDisplayNameText");
   static const userRankCentauriPointsTextKey =
@@ -34,6 +27,13 @@ class RankingCard extends StatelessWidget {
   // This text is used when the parameter [userRank] of the
   // provided [RankingElementDetails] is null.
   static const _nullRankText = "---";
+
+  const RankingCard({
+    super.key,
+    required this.rankingElementDetails,
+  });
+
+  final RankingElementDetails rankingElementDetails;
 
   void onTapUserPopUp(BuildContext context) {
     showDialog(
--- a/lib/views/pages/home/content/ranking/ranking_card.dart
+++ b/lib/views/pages/home/content/ranking/ranking_card.dart
@@ -18,7 +18,7 @@ class RankingCard extends StatelessWidget {
   // Sized box for the rank text to make sure that the text is centered
   // and that surrounding element spacing is consistent.
   // This width is allow to display three digits.
-  static const _sizedBoxRankWidth = 50.0;
+  static const _sizedBoxRankWidth = 35.0;
 
   static const _rankOneColor = Color.fromARGB(255, 255, 218, 75);
   static const _rankSecondColor = Color.fromARGB(255, 217, 218, 204);
@@ -65,9 +65,7 @@ class RankingCard extends StatelessWidget {
           rankingElementDetails.userRank == null
               ? _nullRankText
               : rankingElementDetails.userRank.toString(),
-          style: const TextStyle(
-            fontSize: 24,
-          ),
+          style: Theme.of(context).textTheme.bodyLarge,
         ),
       ),
     );
@@ -78,9 +76,6 @@ class RankingCard extends StatelessWidget {
         key: userRankDisplayNameTextKey,
         rankingElementDetails.userDisplayName,
         overflow: TextOverflow.ellipsis,
-        style: const TextStyle(
-          fontSize: 18,
-        ),
       ),
     );
 
@@ -107,9 +102,6 @@ class RankingCard extends StatelessWidget {
       child: Text(
         key: userRankCentauriPointsTextKey,
         rankingElementDetails.centauriPoints.toString(),
-        style: const TextStyle(
-          fontSize: 18,
-        ),
       ),
     );
 
--- a/lib/views/pages/home/content/ranking/ranking_card.dart
+++ b/lib/views/pages/home/content/ranking/ranking_card.dart
@@ -113,21 +113,23 @@ class RankingCard extends StatelessWidget {
       ),
     );
 
+    final content = InkWell(
+      onTap: () => onTapUserPopUp(),
+      child: Row(
+        children: [
+          userRankText,
+          Expanded(
+            child: userInfo,
+          ),
+          centauriPointsText,
+        ],
+      ),
+    );
+
     return Card(
       clipBehavior: Clip.hardEdge,
       color: cardColor,
-      child: InkWell(
-        onTap: () => onTapUserPopUp(),
-        child: Row(
-          children: [
-            userRankText,
-            Expanded(
-              child: userInfo,
-            ),
-            centauriPointsText,
-          ],
-        ),
-      ),
+      child: content,
     );
   }
 }
--- a/lib/views/pages/home/content/ranking/ranking_card.dart
+++ b/lib/views/pages/home/content/ranking/ranking_card.dart
@@ -73,9 +73,7 @@ class RankingCard extends StatelessWidget {
         child: Text(
           key: userRankrankTextKey,
           textAlign: TextAlign.center,
-          rankingElementDetails.userRank == null
-              ? _nullRankText
-              : rankingElementDetails.userRank.toString(),
+          rankingElementDetails.userRank?.toString() ?? _nullRankText,
           style: themeData.textTheme.bodyLarge,
         ),
       ),
--- a/lib/views/pages/home/content/ranking/ranking_card.dart
+++ b/lib/views/pages/home/content/ranking/ranking_card.dart
@@ -1,0 +1,119 @@
+import "package:flutter/material.dart";
+import "package:proxima/models/ui/ranking/ranking_element_details.dart";
+import "package:proxima/models/ui/user_avatar_details.dart";
+import "package:proxima/views/components/content/publication_header/user_profile_pop_up.dart";
+import "package:proxima/views/components/content/user_avatar/user_avatar.dart";
+
+/// A class that displays a ranking card.
+/// The card displays the user's rank, display name, and centauri points.
+/// The card is clickable and opens a user profile pop-up.
+/// The card's color can be customized.
+class RankingCard extends StatelessWidget {
+  const RankingCard({
+    super.key,
+    required this.rankingElementDetails,
+  });
+
+  // Sized box for the rank text to make sure that the text is centered
+  // and that surrounding element spacing is consistent.
+  // This width is allow to display three digits.
+  static const sizedBoxRankWidth = 50.0;
+
+  static const rankOneColor = Color.fromARGB(255, 255, 218, 75);
+  static const rankSecondColor = Color.fromARGB(255, 217, 218, 204);
+  static const rankThirdColor = Color.fromARGB(255, 229, 153, 137);
+
+  final RankingElementDetails rankingElementDetails;
+
+  void onTapUserPopUp(BuildContext context) {
+    showDialog(
+      context: context,
+      builder: (BuildContext context) => UserProfilePopUp(
+        displayName: rankingElementDetails.userDisplayName,
+        username: rankingElementDetails.userUserName,
+        centauriPoints: rankingElementDetails.centauriPoints,
+      ),
+    );
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final cardColor = switch (rankingElementDetails.userRank) {
+      1 => rankOneColor,
+      2 => rankSecondColor,
+      3 => rankThirdColor,
+      _ => null,
+    };
+
+    final userRankText = Padding(
+      padding: const EdgeInsets.only(left: 10),
+      child: SizedBox(
+        width: sizedBoxRankWidth,
+        child: Text(
+          textAlign: TextAlign.center,
+          rankingElementDetails.userRank == null
+              ? "---"
+              : rankingElementDetails.userRank.toString(),
+          style: const TextStyle(
+            fontSize: 24,
+          ),
+        ),
+      ),
+    );
+
+    final userDisplayNameText = Padding(
+      padding: const EdgeInsets.only(left: 12, right: 16),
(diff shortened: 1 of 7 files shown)


lib/views/pages/home/content/ranking/ranking_card.dart
```diff
@@ -0,0 +1,133 @@
+import "package:flutter/material.dart";
+import "package:proxima/models/ui/ranking/ranking_element_details.dart";
+import "package:proxima/models/ui/user_avatar_details.dart";
+import "package:proxima/views/components/content/publication_header/user_profile_pop_up.dart";
+import "package:proxima/views/components/content/user_avatar/user_avatar.dart";
+
+/// A class that displays a ranking card.
```
gruvw, line 7: a ~class~, a widget

lib/views/pages/home/content/ranking/ranking_card.dart
```diff
@@ -0,0 +1,133 @@
...
+/// The card's color can be customized.
+class RankingCard extends StatelessWidget {
+  const RankingCard({
+    super.key,
+    required this.rankingElementDetails,
+  });
+
+  final RankingElementDetails rankingElementDetails;
+
+  static const userRankrankTextKey = Key("userRankrankText");
```
gruvw, line 19: I usually place `static` `const` at the top of the class so we can clearly see them.

lib/views/pages/home/content/ranking/ranking_card.dart
```diff
@@ -0,0 +1,133 @@
...
+
+  static const _rankOneColor = Color.fromARGB(255, 255, 218, 75);
+  static const _rankSecondColor = Color.fromARGB(255, 217, 218, 204);
+  static const _rankThirdColor = Color.fromARGB(255, 229, 153, 137);
+
+  // This text is used when the parameter [userRank] of the
+  // provided [RankingElementDetails] is null.
+  static const _nullRankText = "---";
+
+  void onTapUserPopUp(BuildContext context) {
```
gruvw, line 38: It is not good practice to pass the `BuildContext` arround like that. This should be a local function, directly defined inside `build`.

lib/views/pages/home/content/ranking/ranking_card.dart
```diff
@@ -0,0 +1,133 @@
...
+        style: const TextStyle(
+          fontSize: 18,
+        ),
+      ),
+    );
+
+    return Card(
+      clipBehavior: Clip.hardEdge,
+      color: cardColor,
+      child: InkWell(
```
gruvw, line 119: Also place this `Inkwell` inside a local `final` variable. Can call it `content` (this is what I usually do).

lib/views/pages/home/content/ranking/ranking_card.dart
gruvw: Same for other documentations in this PR (please apply everywhere).

lib/views/pages/home/content/ranking/ranking_card.dart
gruvw: Same for other `static` `const` in this PR (please apply everywhere).

test/mocks/data/ranking_data.dart
```diff
@@ -0,0 +1,46 @@
+import "package:collection/collection.dart";
+import "package:proxima/models/ui/ranking/ranking_details.dart";
+import "package:proxima/models/ui/ranking/ranking_element_details.dart";
+
+import "firestore_user.dart";
+
+/// Mock user list for the ranking widget.
+final List<Map<String, int>> _rankingUserList = [
```
gruvw, line 8: Good idea to duplicate data here (instead of reusing from UI) so that when integrated with viewmodel this test will still work !

lib/views/pages/home/content/ranking/ranking_widget.dart
```diff
@@ -1,10 +1,75 @@
...
   Widget build(BuildContext context) {
-    return const Center(child: Text("Ranking"));
+    //Mock data creation (should be replaced with actual data)
+    final rankElementDetailsList = userList
+        .mapIndexed(
+          (index, user) => RankingElementDetails(
+            userDisplayName: user.keys.first,
+            userUserName: "u_${user.keys.first}",
+            centauriPoints: user.values.first,
+            userRank: index + 1,
```
gruvw, line 44: The user rank will already be provided with the correct exact rank. I think this should go outside the build method to stay consistent with where you defined your fake data.

Then you should just have a `final rankingDetails` (the one you have below) of type `RankingDetails` inside your build method and you only depend on this variable inside build.
That way when integrating with the viewmodel will come we will simply be able to change the expression of `rankingDetails` inside build using a provider.
Btw you can also add a TODO comment above this `rankingDetails` local variable.

test/views/pages/home/content/ranking/static_ranking_list_test.dart
```diff
@@ -0,0 +1,110 @@
...
+
+      // Check that the centauri points are displayed
+      final centauriPointsTexts =
+          find.byKey(RankingCard.userRankCentauriPointsTextKey);
+      expect(
+        centauriPointsTexts,
+        findsNWidgets(mockRankingElementDetailsList.length),
+      );
+
+      //Check that the card color corresponds to the expected color
```
gruvw, line 65: We should not test for exact match of colors. This test doesn't provide any security as you are not testing external behavior. You just check that what you did is what you did.

gruvw: If you really want to check for UI colors:

A better test would simply check that the first second third and forth element don't all have the same color (without exact math on the color). And that from forth up to last all have the same color.

That way, you only test external behavior.
A general rule of thumb is that a test should not be modified when we simply change a few visual looks (like a color).

gruvw (changes requested):
Great work on this! I left a few remarks (mainly on code quality). Feature wise, it looks exactly like I expected. Can't wait to integrate the viewmodel :smiley:

gruvw:
Also, from a UI/UX note, I feel like the texts are really too large...

If you compare with the feed for example, I think those are the largest texts in the whole application and I don't know if it looks very good.

Maybe we should reduce the font size a little bit, what do you think ?

lib/views/pages/home/content/ranking/ranking_card.dart
CHOOSEIT: You are right !

Fixed in: [b58916a](https://github.com/ProximaEPFL/proxima/pull/322/commits/b58916ab148ea2e6c1db72272e13d2538c2aecf9)

Thank you !

lib/views/pages/home/content/ranking/ranking_card.dart
CHOOSEIT: You are completely right ! It improves readability.

Fixed in: [e156bbc](https://github.com/ProximaEPFL/proxima/pull/322/commits/e156bbc1c67662ccb066c9ab05d4a2a2761f83b6)

Thank you !

lib/views/pages/home/content/ranking/ranking_card.dart
CHOOSEIT: You are right ! That way the code is more readable 💯 

Fixed in: [f805eaf](https://github.com/ProximaEPFL/proxima/pull/322/commits/f805eafbca5da177bb82b1559c1a19f39db0f50d)

Thank you !

lib/views/pages/home/content/ranking/ranking_card.dart
```diff
@@ -0,0 +1,135 @@
...
+  const RankingCard({
+    super.key,
+    required this.rankingElementDetails,
+  });
+
+  final RankingElementDetails rankingElementDetails;
+
+  @override
+  Widget build(BuildContext context) {
+    void onTapUserPopUp() {
```
gruvw, line 40: We are starting to get into the details here, but you don't even need the `void` and you could use the arrow `=>` return syntax here :)

CHOOSEIT:
> Also, from a UI/UX note, I feel like the texts are really too large...
> 
> If you compare with the feed for example, I think those are the largest texts in the whole application and I don't know if it looks very good.
> 
> Maybe we should reduce the font size a little bit, what do you think ?

You are completely right ! I also received the same comment @camillelnne in DM. Also linked to the fact that we could use the theme font size.
I was a bit off in terms of font size compared to the mockup.

I updated this in: [b9b0564](https://github.com/ProximaEPFL/proxima/pull/322/commits/b9b05642a645fc2faceb240584d09eebcfda1329)
I also reduced the `SizedBox` to better fit the mockup.

Tell me what you think 😄

lib/views/pages/home/content/ranking/ranking_card.dart
CHOOSEIT: I like details 👍🏼 
You are right, it improves the readability.

Added in: [8032acc](https://github.com/ProximaEPFL/proxima/pull/322/commits/8032acc554b2bc02e4c74445543b307a88d52f2b)

Thank you !

test/views/pages/home/content/ranking/static_ranking_list_test.dart
CHOOSEIT: I didn't know about this rule of thumb 👍🏼 . Thank you for your comment 💯 

I am going to completely remove the color checking as we may want in the future to modify the color or also add colors to the 4th and 5th places, for instance, or other more complex behavior with colors within the ranking page.

And you are right that it is just a test that verify that the color of the card is the color that I set x)

Removed in: [bcc43d5](https://github.com/ProximaEPFL/proxima/pull/322/commits/bcc43d560ffab941933b3b64ec3210961b11ef0e)

Thank you !

lib/views/pages/home/content/ranking/ranking_widget.dart
CHOOSEIT: You are completely right ! It will make view model integration easier 💯 

Added in: [e482a05](https://github.com/ProximaEPFL/proxima/pull/322/commits/e482a05b8cc8aa25fad04f6804f98b94e3425782) and  [bc9ba87](https://github.com/ProximaEPFL/proxima/pull/322/commits/bc9ba8778fe2a198d85291ac050884e740ad18d0)

Thank you !

lib/views/pages/home/content/ranking/ranking_card.dart
```diff
@@ -0,0 +1,128 @@
...
+    };
+
+    final userRankText = Padding(
+      padding: const EdgeInsets.only(left: 10),
+      child: SizedBox(
+        width: _sizedBoxRankWidth,
+        child: Text(
+          key: userRankrankTextKey,
+          textAlign: TextAlign.center,
+          rankingElementDetails.userRank == null
```
gruvw, line 66: What about `rankingElementDetails.userRank?.toString() ?? _nullRankText` ?

Does it work ?

lib/views/pages/home/content/ranking/ranking_card.dart
```diff
@@ -0,0 +1,128 @@
...
+            context: context,
+            builder: (BuildContext context) => UserProfilePopUp(
+              displayName: rankingElementDetails.userDisplayName,
+              username: rankingElementDetails.userUserName,
+              centauriPoints: rankingElementDetails.centauriPoints,
+            ),
+          ),
+        };
+
+    final cardColor = switch (rankingElementDetails.userRank) {
```
gruvw, line 62: Here I think it would make more sense if we had colors in a `static` `const` list (`[]`) in the order of the rank and then we can query the right color directly using `rankingColors.elementAtOrNull(rankingElementDetails.userRank - 1)`.

What do you think ?

camillelnne (commented):
Hi!

I'm joining this PR because I was super hyped for the feature and tested it on a physical device.
Like I mentioned live to @CHOOSEIT (thanks for the fix!), it's always better to use the theme font, so that style and size remain consistent across the app, ex by defining
Thank you also for the fix of the dark mode, I checked for the colorblindness accessibility and the content is still visible for Deuteranomaly, Protanomaly and Tritanomaly.

I spotted an overflow in case of long pseudo:
WhatsApp Image 2024-05-19 at 14 58 58_327edeb9

But this is problem due to the pop up implementation, so I could create a PR to fix it

test/views/pages/home/content/ranking/static_ranking_list_test.dart
```diff
@@ -0,0 +1,74 @@
...
+import "package:flutter_test/flutter_test.dart";
+import "package:proxima/views/components/content/publication_header/user_profile_pop_up.dart";
+import "package:proxima/views/pages/home/content/ranking/ranking_card.dart";
+import "package:proxima/views/pages/home/content/ranking/ranking_list.dart";
+
+import "../../../../../mocks/data/ranking_data.dart";
+
+void main() {
+  testWidgets(
+    "Check that all the ranking elements are displayed correctly",
```
camillelnne, line 11: I think it would be best to put this test in a group "Widgets display" and name the test with format "Display <elements>" like "Display ranking elements" so that it stays consistent with the rest of UI tests :)

test/views/pages/home/content/ranking/static_ranking_widget_test.dart
```diff
@@ -0,0 +1,39 @@
+import "package:flutter_test/flutter_test.dart";
+import "package:proxima/views/components/content/publication_header/user_profile_pop_up.dart";
+import "package:proxima/views/pages/home/content/ranking/ranking_widget.dart";
+
+import "../../../../../mocks/providers/provider_homepage.dart";
+
+void main() {
+  testWidgets(
+    "Check that all the ranking elements are displayed correctly",
```
camillelnne, line 9: Same for the group and naming of tests here.

CHOOSEIT:
> Like I mentioned live to @CHOOSEIT (thanks for the fix!), it's always better to use the theme font, so that style and size remain consistent across the app, ex by defining Thank you also for the fix of the dark mode, I checked for the colorblindness accessibility and the content is still visible for Deuteranomaly, Protanomaly and Tritanomaly.
> 
> I spotted an overflow in case of long pseudo: 
> But this problem is due to the pop up implementation, so I could create a PR to fix it, if that's okay for you @CHOOSEIT (scrum master).

Thank you for your review, comments and testing. 

Yes this overflow comes from the popUp widget. I think that doing it in a new PR is a good idea 👍🏼 (you can also do it thank you !)

camillelnne (changes requested):
I added a few comments about the name of the tests but you and @gruvw are already doing an awesome job!

I noticed there was a lot of comments and documentation, thank you very much :)

lib/views/pages/home/content/ranking/ranking_card.dart
CHOOSEIT: I think this is a good idea, but since the `userRank` can be null it really complexify the code as we have
```dart
static const _rankingColors = [
    (
      Color.fromARGB(255, 255, 218, 75),
      Color.fromARGB(255, 206, 176, 60)
    ), // Rank 1
    (
      Color.fromARGB(255, 217, 218, 204),
      Color.fromARGB(255, 161, 161, 151)
    ), // Rank 2
    (
      Color.fromARGB(255, 229, 153, 137),
      Color.fromARGB(255, 138, 89, 80)
    ), // Rank 3
  ];

```

But doing that:

```dart 
final cardColorTuple = _rankingColors
        .elementAtOrNull(rankingElementDetails.userRank - 1);
final cardColor = isLightMode ? cardColorTuple?.$1 : cardColorTuple?.$2;
```
does not work as userRank can be null.

So we need to either handle the null case before getting the cardColor. Or doing something like:
```dart 
final cardColorTuple = _rankingColors.elementAtOrNull(
      (rankingElementDetails.userRank ?? _rankingColors.length + 1) - 1,
    );

```

Which I think is not ideal and really degrade the readability of the code.

What do you think ?

lib/views/pages/home/content/ranking/ranking_card.dart
CHOOSEIT: You are right. I think this is a good idea to leverage the operators provided by `Dart`.

Added in: [4c6ca87](https://github.com/ProximaEPFL/proxima/pull/322/commits/4c6ca87510c8f5d470108779e0763fbd54feec58)

Thank you !

lib/views/pages/home/content/ranking/ranking_card.dart
gruvw: What about simply:

```dart
final cardColorTuple = userRank != null ? _rankingColors.elementAtOrNull(userRank - 1) : null;
final cardColor = isLightMode ? cardColorTuple?.$1 : cardColorTuple?.$2;
```

test/views/pages/home/content/ranking/static_ranking_list_test.dart
CHOOSEIT: You are right 💯 

Fixed in: [9f652cc](https://github.com/ProximaEPFL/proxima/pull/322/commits/9f652cc91255a473b8126c34b7d1b479086c7c7a)

Thank you !

test/views/pages/home/content/ranking/static_ranking_widget_test.dart
CHOOSEIT: You are right 💯 

Fixed in: [26c5bbb](https://github.com/ProximaEPFL/proxima/pull/322/commits/26c5bbb2d3238231e450bcfae8c13e16a0eb5bec)

Thank you 😄

lib/views/pages/home/content/ranking/ranking_card.dart
CHOOSEIT: Yeah we can handle the null case with the `?`.

Added in: [d6d832a](https://github.com/ProximaEPFL/proxima/pull/322/commits/d6d832ab7a7422a0dca3d72dbc5aab1bb370ed15)

Thank you !

gruvw (approved):
LGTM now, ready to be merged :rocket: 

**Note**: first merge `ranking` into `main`, then `ranking-ui` into `main`.

camillelnne (approved):
LGTM on my side!

PR #323, opened 2024-05-19
Remove unused UI

### LINKED ISSUE #297: Remove settings button, heatmap chip and badge widget
Acceptance criteria:
- [ ] Remove settings button in profile and in create post page
- [ ] Remove heatmap chip in map page
- [ ] Remove badget widget in profile
- [ ] Update penpot accordingly

camillelnne:
Fixes #297.

This PR removes unused UI elements as part of finalizing the app.

Acceptance criteria:
- [x] Remove settings button in profile and in create post page
- [x] Remove heatmap chip in map page
- [x] Remove badge widget in profile
- [x] Update penpot accordingly. **Note**: After careful consideration, I did not remove any elements because it was decided that we would implement it later, not scrap it completely (Those I had removed were put back in the end).

Additional features:
- Removed the NotImplemented class that we used before we had the ranking.
- Renamed `NavigationbarRoutes` to `NavigationBarRoutes`


lib/views/pages/new_post/new_post_form.dart
```diff
@@ -92,7 +78,12 @@ class NewPostForm extends HookConsumerWidget {
               fit: FlexFit.loose,
               child: _verticallyPadded(bodyField),
             ),
-            _verticallyPadded(buttonRow),
+            _verticallyPadded(
+              SizedBox(
```
JoachimFavre, line 82: This used to be an `Expanded`. By humble curiosity, why did you choose to use a `SizedBox` of infinite width instead?

lib/views/pages/new_post/new_post_form.dart
```diff
@@ -92,7 +78,12 @@ class NewPostForm extends HookConsumerWidget {
...
+                width: double.infinity,
+                child: postButton,
+              ),
```
JoachimFavre, line 85: This used to be an `Expanded`. I'm guessing you decided to use this `SizedBox` for the widget to take the maximum width. 
However, I have to admit I took a bit of time to understand exactly what `width: double.infinity` meant in this scenario. I tried the following that works exactly the same:
```suggestion
              FractionallySizedBox(
                widthFactor: 1.0,
                child: postButton,
              ),
```

What do you think? Does it make the code more readable?

JoachimFavre (commented):
Nice PR! I have one small comment, but it should be quick to resolve! :)

Just as a quick remark, I'm not very fund of 76cd6e1e6666662293ac92faa5f1bb8f9e4f7a17. This was not really part of your sprint task, but it will create merge conflicts with my PRs.
However, since this change is already implemented, I'm guessing it's fine to keep it! :) The conflicts should not be hard to fix at all!

lib/views/pages/new_post/new_post_form.dart
camillelnne: For context, I switched from `Expanded` to `SizedBox` because `Expanded` has to be in a parent widget such as `Row`, which didn't make any sense after we had removed the settings icon.

For `SizedBox` vs `FractionallySizedBox`, I think the opposite of you haha, I feel like `SizedBox` is more straightforward and `FractionallySizedBox` should only be used if we want to set a fraction that is not 1. I think in both cases, it might take a little bit of time for the reader to understand.

Let me know what you think, and we can always ask for a third opinion :)

camillelnne:
Thanks for your review!
I understand your comment about https://github.com/ProximaEPFL/proxima/commit/76cd6e1e6666662293ac92faa5f1bb8f9e4f7a17, I thought because I was already modifying the file (so creating conflicts?), it would be okay to refactor the name, not to add debt (a very minor one in this case) in the backlog.

lib/views/pages/new_post/new_post_form.dart
JoachimFavre: As we discussed on our group chat, it appears @gruvw discussed with a developer on this subject one should use a `Row` with a single `Expanded` child here.

JoachimFavre:
Modifying the file is not enough for creating conflicts. I however ended up merging first, you'll see what I meant about creating conflicts when merging ;)
This should be fast to fix however, don't hesitate if you need any help! :)

lib/views/pages/new_post/new_post_form.dart
camillelnne: Yup, fixed in 44b48a8f55db6932fc8aa9ac5d4e478c27a864a4.

camillelnne:
Merging was fairly easy in the end, just had to fix one name in a test file :)

JoachimFavre (approved):
Yeah, I might have dramatised merging a bit hahahaha. It's nice it was easy to fix! 

All this LGTM, very good job! :)

Issue #302, opened 2024-05-17
Sort posts in profile page


yoannLafore:
Acceptance criteria:
- [ ] Sort the user's posts within the user profile, displaying the latest posts first.

PR #326, opened 2024-05-19
Sort user posts on profile

yoannLafore:
Hello,
This PR aims to sort the user posts on the profile page (issue #302).

Concretelly, this PR does the following:

**User posts view model:**
The view model to display the user posts will now sort the posts from latest to oldest before exposing them.
Thus, the latest posts of the user wil appear at the top of the list on the profile page.

**Mock post firestore generator:**
This PR provides a simple modification to the method `PostFirestoreGenerator::createPostAt`. The modification simply consists in making the mock publication time random. This makes testing more robust and is coherent with the remaining of the method.

**Tests:**
This PR provides a test for the ordering of the user posts exposed by the view model.

#### Live testing
You can test this feature by doing the following:
In the application, add a new post then go into your profile page.
You should then see the newly added post at the top. You can add more posts and you should see them appearing sorted from latest to oldest in the profile page.
sorted_post_profile

#### Acceptance criteria:
- [x] Sort the user's posts within the user profile, displaying the latest posts first.

### DIFF
--- a/lib/viewmodels/user_posts_view_model.dart
+++ b/lib/viewmodels/user_posts_view_model.dart
@@ -21,7 +21,7 @@ class UserPostsViewModel extends AutoDisposeAsyncNotifier<UserPostsState> {
 
     final postsFirestore = await postRepository.getUserPosts(user);
 
-    // Sort the posts by publication time from lattest to oldest
+    // Sort the posts by publication time from latest to oldest
     final sortedPostsFirestore = postsFirestore.toList().sorted(
           (postA, postB) =>
               postB.data.publicationTime.compareTo(postA.data.publicationTime),
--- a/lib/viewmodels/user_posts_view_model.dart
+++ b/lib/viewmodels/user_posts_view_model.dart
@@ -1,4 +1,3 @@
-import "package:collection/collection.dart";
 import "package:hooks_riverpod/hooks_riverpod.dart";
 import "package:proxima/models/database/post/post_id_firestore.dart";
 import "package:proxima/models/ui/user_post_details.dart";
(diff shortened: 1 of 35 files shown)


lib/viewmodels/user_posts_view_model.dart
```diff
@@ -19,7 +20,14 @@ class UserPostsViewModel extends AutoDisposeAsyncNotifier<UserPostsState> {
     final user = ref.watch(validLoggedInUserIdProvider);
 
     final postsFirestore = await postRepository.getUserPosts(user);
-    final posts = postsFirestore.map((post) {
+
+    // Sort the posts by publication time from lattest to oldest
+    final sortedPostsFirestore = postsFirestore.toList().sorted(
+          (postA, postB) =>
+              postB.data.publicationTime.compareTo(postA.data.publicationTime),
```
camillelnne, line 27: What happens if two posts have the same publication ?
For when we will have millions of users x)

lib/viewmodels/user_posts_view_model.dart
```diff
@@ -19,7 +20,14 @@ class UserPostsViewModel extends AutoDisposeAsyncNotifier<UserPostsState> {
     final user = ref.watch(validLoggedInUserIdProvider);
 
     final postsFirestore = await postRepository.getUserPosts(user);
-    final posts = postsFirestore.map((post) {
+
+    // Sort the posts by publication time from lattest to oldest
```
camillelnne, line 24:
```suggestion
    // Sort the posts by publication time from latest to oldest
```
small typo

camillelnne (commented):
Great job as always!

lib/viewmodels/user_posts_view_model.dart
yoannLafore: Good point !

Actually, if two posts have the exact same publication time, the `compareTo` function will return 0.
Thus `postA before postB` and `postB before postA` are both valid ordering.
So, in such cases, multiple valid ordering are possibles.

Note that in practice however, because those posts all belong to the same user, this would mean that the user was able to publish 2 posts within the same microsecond which is unlikely (unless as an adversary behavior).

That was a good question thanks !

lib/viewmodels/user_posts_view_model.dart
yoannLafore: Thanks for the catch :+1: 

Fix in 3b0d4cd8545cfdd067b5599deb827577b66de9b5.

test/viewmodels/user_posts_view_model_integration_test.dart
```diff
@@ -0,0 +1,90 @@
...
+        // Check that the posts list contains exactly one post with the same id
+        final correspondingPosts = posts.where(
+          (post) => post.id == userPostDetails.postId,
+        );
+        expect(correspondingPosts, hasLength(1));
+
+        return correspondingPosts.first;
+      }
+
+      test("User Posts are ordered from latest to oldest", () async {
```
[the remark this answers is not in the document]
yoannLafore: That is a good idea !

I add tests for those edge cases in 42c1600a84ab311ef09cad23f12d73e9a08178d9 and 02e00422a389f030b6162b66f51fedc5d9e2def9.

lib/viewmodels/user_posts_view_model.dart
camillelnne: Oh yes, that makes sense, thanks for the reply!

camillelnne (approved):
Thanks for the changes!
I hope the old comments erasing won't be too hard.

yoannLafore:
Thank you for all your great suggestions :tada:

PR #330, opened 2024-05-19
Fix overflow user pop up

### LINKED ISSUE #327: Bug fix: Profile popup overflow
Acceptance criteria:
- [ ] Fix text overflow on profile popup on:
    - [ ] Displayname
    - [ ] Username
    - [ ] Centauri points

camillelnne:
Fixes #327.
This PR fixes the overflow in the user pop up caused by a long pseudo.

image

**Features**:
- Ellipsis on pseudo overflow in pop up
- Reduction of the maximum number of character for username and pseudo after team decision.
- Centauri points and username overflow are already handled by wrapping

**Note**:  PublicationHeader, was placed in `views/components/` and its name was not consistent with our `PostCard` component which made it difficult for me to find when starting to work on this PR, so I moved and renamed it. 
I understand that it was not in the scope of the PR, so I leave the choice to the reviewer to have it reverted or not. If kept, this PR will be merged at the end of the sprint not to hinder any other teammate progression.

**Acceptance criteria:**
Fix text overflow on profile popup on:
- [x] Displayname
- [x]  Username
- [x] Centauri points


lib/viewmodels/create_account_view_model.dart
```diff
@@ -31,18 +34,18 @@ class CreateAccountViewModel extends AsyncNotifier<CreateAccountValidation> {
...
       return "Too short.";
     }
 
-    if (value.length > 20) {
+    if (value.length > _maxStringLength) {
       return "Too long.";
     }
 
     // This regular expression is intentionally too restrictive (for instance,
     // the {3,20} is already checked above). Its purpose is too make sure that
```
TheTexanCodeur, line 46:
```suggestion
    // the {_minStringLength,_maxStringLength} is already checked above). Its purpose is too make sure that
```

test/views/pages/create_account/create_account_test.dart
```diff
@@ -53,7 +53,7 @@ void main() {
       ("heLlo ", space),
       ("A", short),
       ("a%", short),
-      ("0" * 21, long),
+      ("0" * 17, long),
```
TheTexanCodeur, line 56: You created the _minStringLength and _maxStringLength variables in [lib/viewmodels/create_account_view_model.dart](https://github.com/ProximaEPFL/proxima/pull/330/files#diff-97ada99179c9b08c7aad6f9b10bb328c8b43fba692d917e3f167455dcb886e42). Do you think it would be a good idea to do the same thing in this test file? It would have the same purpose (avoid hardcoded values)

TheTexanCodeur (commented):
Good job! 

This bug fix / refactor make the app safer and the code easier to read. 
I just made minor suggestions. This PR is almost ready to merge 👍🏼

lib/viewmodels/create_account_view_model.dart
camillelnne: Good catch! Fixed in 083d1415676d38fe2fe3537cfd43208c7a280448

test/views/pages/create_account/create_account_test.dart
camillelnne: Yes, super good idea! I actually struggled updating the tests because of the magic numbers so it would fix the problem.

camillelnne: fixed in 2612b02017b1ecfdcd5fcf953ac01aa6eace7008

TheTexanCodeur (approved):
Thanks for taking my suggestions into account.

LGTM! 🔥

PR #334, opened 2024-05-21
Challenge refresh page + refresh page refactor

### LINKED ISSUE #301: Bug: challenge refresh page after error
The challenge page should have a refresh button page similar to the feed page when there is an error. Right now it is just empty.

Close #186 if decide to do that.

Acceptance criteria:
- [ ] : it is possible to refresh the challenge page if there is an error

Aderfish:
Fixes #301 and #186,

The challenge page now shows an error text and a refresh button after an error.

I did some refactoring on the feed and map view, by putting the duplicated code into the `ErrorRefreshPage` widget so that we can easily reuse it.

Acceptance criteria:
- [x] : it is possible to refresh the challenge page if there is an error


lib/views/pages/home/content/map/map_screen.dart
```diff
@@ -52,7 +34,9 @@ class MapScreen extends ConsumerWidget {
         );
       },
       fallbackBuilder: (context, error) {
-        return fallback;
+        return ErrorRefreshPage(
+          onRefresh: () => ref.read(mapViewModelProvider.notifier).refresh(),
```
TheTexanCodeur, line 38: Here I think that we could declare a final function above like you did for the rest of the refactor.

TheTexanCodeur (commented):
Good job!

The refactor makes the code cleaner and the new tests are easy to read/understand.
I just made a little suggestion to ensure that the refactor is consistent. Other than that this PR looks ready to merge 👍🏼

lib/views/pages/home/content/map/map_screen.dart
Aderfish: I added a variable in the other screens because the function was reused multiple times, why do you think I need to add one here ?

TheTexanCodeur (approved):
Thanks for taking my suggestions into account!

LGTM 🔥

lib/views/pages/home/content/map/map_screen.dart
TheTexanCodeur: You're completely right, I didn't think about that!

PR #336, opened 2024-05-23
More end to end tests

### LINKED ISSUE #220: Extend end-to-end tests for challenges and map
Acceptance criteria: TBD after finishing user stories
Could be split in two tasks, considering that some devs might be more efficient for some tasks.

Aderfish:
Fixes #220,

I have added end to end tests for completing a challenge and leaving and deleting a comment. I've had to fix a bug in the `createUserPost` test util to make the geohash of the generated post correct.

Acceptance criteria:
- [x] : challenge generation and completion
- [x] : commenting on a post
- [x] challenges update on user position change


test/mocks/data/geopoint.dart
```diff
@@ -101,4 +101,22 @@ class GeoPointGenerator {
...
+  ) {
+    double latStep = (end.latitude - start.latitude) / steps;
+    double lonStep = (end.longitude - start.longitude) / steps;
+
+    return List.generate(steps + 1, (i) {
+      return GeoPoint(
+        start.latitude + latStep * i,
+        start.longitude + lonStep * i,
+      );
+    });
```
yoannLafore, line 120: Good idea to add this function !
I just believe we can make it a little bit more concise using the arrow synthax here:

```suggestion
    return List.generate(
      steps + 1,
      (i) => GeoPoint(
        start.latitude + latStep * i,
        start.longitude + lonStep * i,
      ),
    );```

File: test/end2end/app_test.dart [hunk 2]
```diff
@@ -64,6 +78,15 @@ void main() {
     await tester.pumpAndSettle();
   }
 
+  void goToPoint(GeoPoint point) {
```
yoannLafore, line 81: I just wonder if we could add a little line of documentation here.

test/end2end/app_test.dart
```diff
@@ -72,9 +95,177 @@ void main() {
...
+
+  testWidgets("Challenge creation and completion", (WidgetTester tester) async {
+    // create a post that will be the challenge
+    final otherUser = await FirestoreUserGenerator.addUser(fakeFireStore);
+    final postLocation = GeoPointGenerator.createOnEdgeInsidePosition(
+      startLocation,
+      ChallengeRepositoryService.maxChallengeRadius,
+    );
+    final post =
+        FirestorePostGenerator().createUserPost(otherUser.uid, postLocation);
```
yoannLafore, line 110: Good jov with this test :+1: 
I just think we could instanciate the `FirestorePostGenerator` in the setup like that we don't need to re-instanciate it in each test.
What do you think ?

yoannLafore (commented):
Good job !
The added end2end tests are complete and robustly test the challenges and commenting user flow.

The code is well moduralised with a lot of helper functions :+1: 
I was simply wondering if we should go one step further and maybe create a utility static class for end2end tests with all those function (`expectCommentCount`, `addComment`, `openPost`, ...) ?
I am not entirely sure however and I'd like to get to point of view on that.

Other than that the PR looks good to me.
I have just made a few style suggestions.

Thanks for your work !
Great job :tada:

test/end2end/app_test.dart
Aderfish: feb1b7e529e2a4dd9ef704dd89ea84a3ad25657c

test/end2end/app_test.dart
Aderfish: 1df14d23b2b0e155c512a111bfc91c5b3f36c958

Aderfish:
I think separating the helper functions is a good idea, and that way we might reuse them in other places. cb8d423967f3c320a4dc2912c48105ce08b72ced

yoannLafore (commented):
I have just have a last suggestion below and other than that the PR looks good to me.

test/end2end/actions.dart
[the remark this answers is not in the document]
Aderfish: That makes sense, done in 6a4e6f81e51dd39c830bbf65c8aac5ff772ef98a

yoannLafore (approved):
Thanks for the changes !
The PR looks good to me now :+1: 

Amazing job :tada:

PR #338, opened 2024-05-24
Check that tag does not already exist on release

gruvw:
Adds a check on the CI release pipeline to avoid overriding a previous release in case the workflow is triggered without a new (bumped) app version.

**Note**: workflow tested in a fork.


camillelnne (approved):
LGTM

Issue #342, opened 2024-05-24
Bug: Avatar color in app bar doesn't update on challenges completed


Aderfish:
Acceptance criteria:
- [ ] Make the user avatar color update when completing a challenge

![Image](https://github.com/ProximaEPFL/proxima/assets/101193378/950784eb-bfa6-47ef-ba35-334627c098ca)

Issue #316, opened 2024-05-18
Refresh comment count on navigation back


CHOOSEIT:
Acceptance criteria:
- When commenting on the post page and going back to feed, the number of comments on post card should be updated

PR #348, opened 2024-05-25
Bug fix: user avatar does not update on challenge completion

CHOOSEIT:
Fixes: #342 

This PR aims to fix the user avatar not updating properly on challenge completion.

image

As you can see with the line changes, this was not a straight-forward bug fix. 
We had not any updates on the user avatars. The user avatars were also not linked to any provider, hence, on challenge completion, we had no way to update every user avatar at once.

This PR creates a new provider `userCentauriPointsViewModelProvider`, that gets the centauri points of a user given its user id. This provider is refreshable, allowing us to watch on it in every user avatar.

Note that the user avatar had not access to the user id of the displayed user, so a considerable part of this PR is giving this access to this information.


test/models/ui/comment_details_test.dart
```diff
@@ -47,5 +48,12 @@ void main() {
 
       expect(commentPost != commentPostOther, true);
     });
+
+    test("inequality test on owner ID", () {
+      final commentPost = unequalComments[0];
+      final commentPostOther = unequalComments[4];
+
+      expect(commentPost != commentPostOther, true);
```
yoannLafore, line 56: Very small suggestion, I found out that we could test inequality with the matchers `isNot(equals(***))`.
This would gives:
```suggestion
      expect(commentPost, isNot(equals(commentPostOther)));
```

I personnaly feels like it improves readability a little bit.
But this is just my point of view and I'd totally understand if you choose to stick with your approach :)

test/models/ui/ranking/ranking_element_details_test.dart
```diff
@@ -83,18 +92,40 @@ void main() {
...
+
+      final rankingElementDetail2 = RankingElementDetails(
+        userDisplayName: testingUserData.displayName,
+        userUserName: testingUserData.username,
+        userID: const UserIdFirestore(value: "not_the_same_uid"),
+        centauriPoints: 10,
+        userRank: 1,
+      );
+
+      expect(rankingElementDetail1 == rankingElementDetail2, isFalse);
```
yoannLafore, line 128: If you choose to go with my previous suggestion, I think it could also be applied here.

```suggestion
      expect(rankingElementDetail1, isNot(equals(rankingElementDetail2)));
```

test/viewmodels/user_centauri_points_view_model_unit_test.dart
```diff
@@ -0,0 +1,63 @@
...
+      await setUsersFirestore(fakeFireStore, availableUsers);
+
+      //Prepare the list of providers to override.
+      container = ProviderContainer(
+        overrides: [
+          firestoreProvider.overrideWithValue(fakeFireStore),
+          userRepositoryServiceProvider.overrideWithValue(userRepo),
+        ],
+      );
+    });
```
yoannLafore, line 37: I think `userRepo` isn't use in the tests. 
Thus I believe we can remove it.

```suggestion
    late FakeFirebaseFirestore fakeFireStore;

    late ProviderContainer container;

    late List<UserFirestore> availableUsers;

    setUp(() async {
      fakeFireStore = FakeFirebaseFirestore();

      availableUsers = FirestoreUserGenerator.generateUserFirestore(10);
      await setUsersFirestore(fakeFireStore, availableUsers);

      //Prepare the list of providers to override.
      container = ProviderContainer(
        overrides: [
          firestoreProvider.overrideWithValue(fakeFireStore),
        ],
      );
    });
```

yoannLafore (commented):
Thank you for fixing this bug !
This effectively contributes to making our application more responsive to the user and so improves the overall UX.

Your code is of great quality with abondant documentation :+1: 
The PR overall looks good to me and I saw that you added some tests for the new view model which is great.

I have just made a few suggestion below.

Thanks for your great work :tada:

lib/utils/user_avatar_color.dart
```diff
@@ -0,0 +1,50 @@
+import "package:flutter/material.dart";
+import "package:proxima/models/ui/linear_segmented_hsv_colormap.dart";
+import "package:proxima/services/database/challenge_repository_service.dart";
+
+/// Class that provides the color the user's avatar
+/// based on their centauri points.
+class UserAvatarColor {
```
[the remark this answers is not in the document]
gruvw: btw in dart I think we should avoid class with only static members, check out https://dart.dev/tools/linter-rules/avoid_classes_with_only_static_members

lib/models/ui/user_avatar_details.dart
```diff
@@ -1,42 +1,47 @@
...
-    required this.userCentauriPoints,
+    required this.userID,
   });
 
   /// Converts a [UserData] object, [userData], to a [UserAvatarDetails] object.
-  factory UserAvatarDetails.fromUserData(UserData userData) {
+  factory UserAvatarDetails.fromUserData(
+    UserData userData,
+    UserIdFirestore userID,
+  ) {
```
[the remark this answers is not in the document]
CHOOSEIT: You are completely right ! 

Added in: [7a9eeaf](https://github.com/ProximaEPFL/proxima/pull/348/commits/7a9eeafec0ee9dfe285594e4d73d03bdfd407de1)

Thank you !

lib/utils/user_avatar_color.dart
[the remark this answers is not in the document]
CHOOSEIT: You are completely right 💯 

I removed the class in : [4d15cbb](https://github.com/ProximaEPFL/proxima/pull/348/commits/4d15cbbb38197d4aae52aeb6720e5c40d40f3835)

Thank you !

lib/viewmodels/dynamic_user_avatar_view_model.dart
```diff
@@ -17,65 +14,29 @@ class DynamicUserAvatarViewModel extends AutoDisposeFamilyAsyncNotifier<
...
+      userID = arg;
     }
 
-    return UserAvatarDetails.fromUserData(user.data);
+    user = await userDataBase.getUser(userID);
+
+    return UserAvatarDetails.fromUserData(
+      user.data,
+      userID,
+    );
```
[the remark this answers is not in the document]
CHOOSEIT: Yes, that's clearly more readable 😄 

Changed also in: [7a9eeaf](https://github.com/ProximaEPFL/proxima/pull/348/commits/7a9eeafec0ee9dfe285594e4d73d03bdfd407de1)

Thank you !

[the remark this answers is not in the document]
CHOOSEIT: Yes, that's more concise 😄 

Changed in: [7a9eeaf](https://github.com/ProximaEPFL/proxima/pull/348/commits/7a9eeafec0ee9dfe285594e4d73d03bdfd407de1)

Thank you !

lib/viewmodels/user_centauri_points_view_model.dart
```diff
@@ -0,0 +1,42 @@
...
+
+  /// Refresh the number of centauri points of the user.
+  Future<void> refresh() async {
+    state = await AsyncValue.guard(() => build(arg));
+  }
+
+  /// Refresh the number of centauri points of the user.
+  Future<void> refreshWithCentauriPointsNumber(int centauriPoints) async {
+    state = AsyncValue.data(centauriPoints);
+  }
```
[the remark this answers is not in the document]
CHOOSEIT: You are right ! It can create some confusion when using it.

Changed in: [f5fc5a7](https://github.com/ProximaEPFL/proxima/pull/348/commits/f5fc5a75ada6f66e41447466e0397ce4fa6c98f5)

Thank you !

test/models/ui/ranking/ranking_element_details_test.dart
CHOOSEIT: You are completely right.

Changed in [cefc851](https://github.com/ProximaEPFL/proxima/pull/348/commits/cefc85186f44b65ca6bdfe9f097aaea631128868)

I also added this matcher to the other tests in this file :)

Thank you !

test/models/ui/comment_details_test.dart
CHOOSEIT: You are completely right.

Changed in [cefc851](https://github.com/ProximaEPFL/proxima/pull/348/commits/cefc85186f44b65ca6bdfe9f097aaea631128868)

I also added this matcher to the other tests in this file :)

Thank you !

test/models/ui/ranking/ranking_element_details_test.dart
CHOOSEIT: You are right !

Applied in: [c7be8de](https://github.com/ProximaEPFL/proxima/pull/348/commits/c7be8de3f35c0cf986dac822e0e5fe83c1b4c4b1)

Note that I also applied this change for the other tests in the file :)

Thank you !

test/viewmodels/user_centauri_points_view_model_unit_test.dart
CHOOSEIT: You are completely right !

Removed in [7126752](https://github.com/ProximaEPFL/proxima/pull/348/commits/7126752cd987a85b2aa4aeac237f6501bbe51336)

Thank you !

lib/utils/user_avatar_color.dart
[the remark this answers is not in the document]
yoannLafore: I didn't know that either, thanks for pointing it out !

Also, thanks for making the change :+1:

test/models/ui/comment_details_test.dart
yoannLafore: Perfect :ok_hand: 
Thank you !

yoannLafore (approved):
The PR looks good to me now :+1: 

Amazing job :tada:

CHOOSEIT:
Thank you, @yoannLafore, for your review 🎉 

As this PR changes a lot of things, we are going to wait for @JoachimFavre's review before merging 💯

lib/viewmodels/users_ranking_view_model.dart
```diff
@@ -64,6 +68,24 @@ class UsersRankingViewModel extends AutoDisposeAsyncNotifier<RankingDetails> {
...
+            topUsers.data.centauriPoints,
+          );
+    }
+
+    //Refresh the centauri points of the current user
+    ref
+        .read(
+          userCentauriPointsViewModelProvider(currentUser.uid).notifier,
+        )
+        .refreshWithCentauriPointsNumber(currentUserData.centauriPoints);
```
JoachimFavre, line 87: The following allows to slightly decrease code repetition:
```
    // Refresh the centauri points of the top users and the current user
    for (final user in topUsersFromDb + [currentUser]) {
      ref
          .read(
            userCentauriPointsViewModelProvider(user.uid).notifier,
          )
          .refreshWithCentauriPointsNumber(
            user.data.centauriPoints,
          );
    }
```

This is not a GitHub suggestion since I think it requires removing unused variables.

lib/utils/user_avatar_color.dart
```diff
@@ -0,0 +1,46 @@
+import "package:flutter/material.dart";
+import "package:proxima/models/ui/linear_segmented_hsv_colormap.dart";
+import "package:proxima/services/database/challenge_repository_service.dart";
+
+/// Stops for the colormap used to color the user's avatar based on their centauri points.
```
JoachimFavre, line 5: By humble curiosity, why did you do this refactor?

I'm not a big fan of its current form, because I feel like it adds some namespace pollution. If you don't want to put those lines back to the dynamic user avatar view-model, I think you should at least encapsulate everything inside a class `UserAvatarColor` with a private constructor.

lib/views/components/content/user_avatar/user_avatar.dart
```diff
@@ -25,35 +28,53 @@ class UserAvatar extends StatelessWidget {
...
+          child: content,
+        );
+      },
+      loading: () => CircleAvatar(
+        radius: radius,
+        backgroundColor: loadingUserAvatarColor,
+        child: content,
+      ),
+      error: (error, _) => throw Exception(
+        "${CircularValue.debugErrorTag} User avatar color error: $error",
```
JoachimFavre, line 77: You can make this code much simpler by taking advantage of `AsyncValue::value` which returns the previous value if it exists or null otherwise, and re-throws the error if the `AsyncValue` is in an error state; and with the fact that `centauriToUserAvatarColor` returns the loading value when the number of points given in argument is `null`.
```
    final backgroundColor = centauriToUserAvatarColor(
      userCentauriAsync.value,
      Theme.of(context).brightness,
    );

    return CircleAvatar(
      radius: radius,
      backgroundColor: backgroundColor,
      child: content,
    );
```

In fact, using this method, you can go back to the way this file originally was by adding very few lines.

lib/viewmodels/user_centauri_points_view_model.dart
```diff
@@ -0,0 +1,41 @@
...
+    final userID = arg;
+
+    final user = await userDataBase.getUser(userID);
+    final userCentauri = user.data.centauriPoints;
+    return userCentauri;
+  }
+
+  /// Refresh the number of centauri points of the user.
+  Future<void> refresh() async {
+    state = await AsyncValue.guard(() => build(arg));
```
JoachimFavre, line 30: If you decide to apply my previous comment, I'd recommend to also use the following here:
```suggestion
    state = const AsyncValue.loading();
    state = await AsyncValue.guard(() => build(arg));
```

This allows to let the UI decide whether it want to show the user the value is loading, or instead pick the last existing value.

lib/viewmodels/user_centauri_points_view_model.dart
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  /// Refresh the number of centauri points of the user.
+  void refreshWithCentauriPointsNumber(int centauriPoints) =>
+      state = AsyncValue.data(centauriPoints);
```
JoachimFavre, line 35: After a discussion with @gruvw on a similar matter for my PR [#316](https://github.com/ProximaEPFL/proxima/issues/316), it appears that we may want to be careful with such small optimisations over Firestore's cache. It is probably better to be consistent everywhere, i.e. to trust the cache or to make optimisations everywhere.

test/mocks/data/post_comment.dart
```diff
@@ -66,6 +74,7 @@ final List<CommentDetails> unequalComments = [
     content: "content1",
     ownerDisplayName: "username2",
     ownerUsername: "username2username",
+    ownerUserID: const UserIdFirestore(value: "username2ID"),
```
JoachimFavre, line 77: To be consitent with the comment, we should have:
```suggestion
    ownerUsername: "username1username",
    ownerUserID: const UserIdFirestore(value: "username1ID"),
```

test/mocks/overrides/override_user_centauri_points_view_model.dart
```diff
@@ -0,0 +1,47 @@
...
+
+  @override
+  Future<void> refresh() => _onRefresh();
+
+  @override
+  void refreshWithCentauriPointsNumber(int centauriPoints) =>
+      _onRefreshWithCentauriPointsNumber();
+}
+
+List<Override> mockUserCentauriPointsViewModelCentauriOverride({
```
JoachimFavre, line 34: I think it is worth adding some documentation here

test/mocks/providers/provider_ranking.dart
```diff
@@ -1,40 +1,50 @@
...
   FakeFirebaseFirestore fakeFirebaseFirestore,
 ) async {
   final rankingContainer =
       await rankingProviderContainerWithTestingUser(fakeFirebaseFirestore);
 
-  return MaterialApp(
-    home: UncontrolledProviderScope(
-      container: rankingContainer,
-      child: const RankingPage(),
+  return ProviderScope(
```
JoachimFavre, line 42: By humble curiosity, why did you wrapped this `MaterialApp` in a `ProviderScope`?

JoachimFavre (changes requested):
Good job with this great PR! I have some comments, don't be afraid of their number their number is simply proportional to the size of your PR (and, they should be quick to fix!)

I just have another general comment. I feel like you no longer a `DynamicUserAvatar` and a `UserAvatar`, we can get rid of `DyanmicUserAvatar, can't we?

lib/utils/user_avatar_color.dart
gruvw: please check previous reviews, notably https://github.com/ProximaEPFL/proxima/pull/348#discussion_r1614739378, I think it will answer your question

CHOOSEIT: I think it does not make sense to leave it in the dynamic user avatar as it does not handle color anymore.

I think this is linked to this suggestion: https://github.com/ProximaEPFL/proxima/pull/348#discussion_r1614739378

lib/viewmodels/users_ranking_view_model.dart
CHOOSEIT: You are right !

Added in: [3347ba7](https://github.com/ProximaEPFL/proxima/pull/348/commits/3347ba7256a1066d388c2e66814259b6399a4be8)

Thank you !

lib/viewmodels/user_centauri_points_view_model.dart
CHOOSEIT: I did not know 👍🏼 

I decided to remove in [243304f](https://github.com/ProximaEPFL/proxima/pull/348/commits/243304f86b9f757c4460da135fb1ba639f882e5f) the `.refreshWithCentauriPointsNumber` as the` .refresh` have a more general usage.

Thank you 💯 

Tell me what you think 👍🏼

test/mocks/data/post_comment.dart
CHOOSEIT: You are right! 

Fixed in :[18b20c1](https://github.com/ProximaEPFL/proxima/pull/348/commits/18b20c1ce2b4d53f54fcebdda746e1683f3668a8)

Thank you 💯

test/mocks/providers/provider_ranking.dart
CHOOSEIT: A test was failing on a pop-up that was not finding the `ProviderScope`.

Thanks to you, I came back on this, and realised that we already had a `ProviderScope` but that was not wrapping the `MaterialApp` as we usually do, that is why I wrapped it.

Fixed in: [6b09986](https://github.com/ProximaEPFL/proxima/pull/348/commits/6b0998610759ac85c5dc49153e8ebe78e2f1ed05)

Thank you 💯

test/mocks/overrides/override_user_centauri_points_view_model.dart
CHOOSEIT: You are right 💯 

Added in: [54b80e6](https://github.com/ProximaEPFL/proxima/pull/348/commits/54b80e62e6891282a6cd1a744d8a32d5e7c209a1)

Thank you !

lib/views/components/content/user_avatar/user_avatar.dart
CHOOSEIT: You are right.

Fixed in: [ad5cdda](https://github.com/ProximaEPFL/proxima/pull/348/commits/ad5cddab335c641e4c227995e7a579559f0aaf19)

Thank you 💯

lib/viewmodels/user_centauri_points_view_model.dart
CHOOSEIT: You are right ! 

Added in: [5154d96](https://github.com/ProximaEPFL/proxima/pull/348/commits/5154d966387b81f654a70a210edc41deec765ae3)

Thank you !

lib/viewmodels/user_centauri_points_view_model.dart
```diff
@@ -0,0 +1,41 @@
...
+    extends FamilyAsyncNotifier<int?, UserIdFirestore?> {
+  UserCentauriPointsViewModel();
+
+  @override
+  Future<int?> build(UserIdFirestore? arg) async {
+    final userDataBase = ref.watch(userRepositoryServiceProvider);
+
+    if (arg == null) {
+      return null;
+    }
```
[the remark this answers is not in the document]
CHOOSEIT: You are completely right, however I think it relies on doing a conditional `watch` in the UI with the current state of the code, which is not desired. I think that doing a clean solution is much wider than this PR's scope and should be addressed in a separate PR, along with the potential merge of `DynamicUserAvatar` and `UserAvatar` and the removal of nullable user IDs. 👍🏼 

Tell me what you think: 💯

CHOOSEIT:
> I just have another general comment. I feel like you no longer a `DynamicUserAvatar` and a `UserAvatar`, we can get rid of `DyanmicUserAvatar`, can't we?

You are completely right ! I answered in this comment: https://github.com/ProximaEPFL/proxima/pull/348#discussion_r1615254059

lib/viewmodels/user_centauri_points_view_model.dart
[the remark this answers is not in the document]
JoachimFavre: Mhm I see indeed. This would require a big refactor to never pass a null id; which is not a choice you took in this PR. This makes sense ok, you can stay like this.

lib/utils/user_avatar_color.dart
JoachimFavre: Hahahah! I read Yoann's comments and it appears I have managed to miss this one and only comment. That's sad x)

Thank you, I learnt something today :D

lib/viewmodels/user_centauri_points_view_model.dart
JoachimFavre: This is perfect :)

(Don't worry, I did not know this morning either haha)

lib/views/components/content/user_avatar/user_avatar.dart
JoachimFavre: Thanks. However, why did you choose to keep the exception re-throw? I feel like it does not bring much to the code, but still increases slightly the cognitive complexity while destroying the stack trace of the original error (preventing us from knowing from where it originated).

test/mocks/providers/provider_ranking.dart
JoachimFavre: Ah yes, I understand better now! Great, thank you 🚀

lib/views/components/content/user_avatar/user_avatar.dart
CHOOSEIT: Oh, I see what you mean 😄 . Moreover since we don't always have a circular value backing the the widget, is the not relevant and this is better to directly throw the error without breaking the stack trace.

Removed in: [7850437](https://github.com/ProximaEPFL/proxima/pull/348/commits/7850437df297d31ee796897e115b38325b7682e3)

Thank you 💯

lib/viewmodels/user_centauri_points_view_model.dart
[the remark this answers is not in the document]
CHOOSEIT: I agree. 

Modified in: [0e38ef4](https://github.com/ProximaEPFL/proxima/pull/348/commits/0e38ef4bcd244327ec37406349edaacc520318b2)

Thank you ! 

Note that I made many changes, let us continue this in the main thread as I am going to ask @yoannLafore to re-review as a big chunk changed 👍🏼

CHOOSEIT:
As I changed a big chunk since the approval of @yoannLafore, I think he should take a look at the new code.

lib/viewmodels/user_centauri_points_view_model.dart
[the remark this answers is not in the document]
CHOOSEIT: Note that I added the task of accepting a non-nullable `userId` in the dynamic user avatar view model: [#353](https://github.com/ProximaEPFL/proxima/issues/353)

yoannLafore (approved):
I have gone through the resolved conversations and they look good to me :+1:
I also learned a few things that I didn't know about dart.

Thanks for your great work @CHOOSEIT and @JoachimFavre :tada: 
The PR looks good to me.

lib/viewmodels/user_centauri_points_view_model.dart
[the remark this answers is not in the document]
JoachimFavre: You refactor is perfect, great job! :)

Yes, I agree with the fact that creating a new task to the backlog for the `userId` is the good choice, you've already done a lot in this PR; and this was not a choice you took in this PR. Great job :)

JoachimFavre (approved):
You made an amazing job at fixing my comments, I don't have anything to say on the code you modified. Great job!

I'm sorry for the refactor to move the points refresh to the dynamic profile view-model, I have to admit I did not really anticipate it would be so big. I however find your code absolutely incredible after this refactor, so thank you so much for doing it! :)

LGTM!

PR #349, opened 2024-05-25
Update comment count when adding a comment

### LINKED ISSUE #316: Refresh comment count on navigation back

JoachimFavre:
Fixes #316.

This PR adds a `PostCommentCountViewModel`, which stores a post number of comment (in a similar fashion to `PostVotesViewModel`). It is refreshed by doing a fetch to the database when a comment is deleted, and refreshed efficiently when the comment list is refreshed on the post page.

This is then tested by taking advantage of the great work @Aderfish made for end-to-end tests. 

Happy reviewing! :)

Acceptance criteria:
- [x] When commenting on the post page and going back to feed, the number of comments on post card should be updated


lib/viewmodels/post_comment_count_view_model.dart
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  /// Set the count of comments of the post to [count]. This updates
+  /// the state without needing to fetch the data from the database.
+  void setCount(int count) {
+    state = AsyncValue.data(count);
+  }
+
+  Future<void> refresh() async {
+    state = await AsyncValue.guard(() => build(arg));
```
gruvw, line 28: Is it voluntary to have no loading state here ?

Refreshing a viewmodel is always done using the following:
```dart
  Future<void> refresh() async {
    state = const AsyncValue.loading();
    state = await AsyncValue.guard(() => build()); // with or without arg
  }
```

If yes why ?

If the goal here is to avoid the loading in the UI, that should be done directly in the UI, but the viewmodel should still have a loading state.

lib/viewmodels/comments_view_model.dart
```diff
@@ -45,6 +46,9 @@ class CommentsViewModel extends AutoDisposeFamilyAsyncNotifier<
           -commentA.publicationDate.compareTo(commentB.publicationDate),
     );
 
+    // Update the post comment count
+    ref.read(postCommentCountProvider(arg).notifier).setCount(comments.length);
```
gruvw, line 50: I am not sure why you call `setCount` here, and about the relevance of the `setCount` altogether.

Why do you need to set the precise count "without fetching the data" ?
I think Firebase already does caching and such optimizations, so we should simply call refresh to get the new data back.
Did you test this and found out it has measurable performance implications ?
I don't see any other viewmodel in the codebase where we applied such shortcuts of the database to avoid making a (probably local) Firebase query.

test/end2end/app_actions.dart
gruvw, line 1: It's interesting to have performed all the testing of this feature only in end2end tests.

I think it is ok, as we are in the last sprint of the last milestone (and we have more than enough coverage) we probably don't need a special test that isolates this new behavior.

gruvw (changes requested):
Good code with extensive documentation. Left a few remarks, but it should be ready to merge soon.

lib/viewmodels/comments_view_model.dart
gruvw: You mentioned the `PostVotesViewModel` in the PR description but there it was not to be more efficient, it was from a user perspective to avoid directly showing a larger number change.

The comment count is not updated/visible directly when the user performs an action that could impact its value.
Also when the user posts a new comment we refetch all the comments, so the count will be updated with the actual value of the new number of comments in the database.

lib/viewmodels/user_comments_view_model.dart
```diff
@@ -53,6 +54,8 @@ class UserCommentViewModel extends AutoDisposeAsyncNotifier<UserCommentsState> {
     // Refresh the home feed after comment deletion, that way the comment
     // count will be updated
     ref.read(postsFeedViewModelProvider.notifier).refresh();
+    // Also update the comment count for the post
+    ref.read(postCommentCountProvider(postId).notifier).refresh();
```
gruvw, line 58: You also mention this in the PR description, but why did you choose to refresh the count differently on comment deletion vs on refresh of a post's comments ?

lib/viewmodels/post_comment_count_view_model.dart
JoachimFavre: It was voluntary. This came because I did not understand `AsyncValue::value` and `AsyncValue::valueOrNull` correctly. Now that I do, this was fixed in 810bf0977a466278e0921c81264e6ec3baea86e8. I moreover decided to use `value` instead of `valueOrNull` to propagate potential errors arrising in the `AsyncValue` in 70ac3f7c443984b6f99dba6c0578a805ac0d7b9a.

lib/viewmodels/comments_view_model.dart
JoachimFavre: My intuition was that the firestore cache could not be fully trusted because, when we go offline, a fetch to the database takes a time hinting that it tried to make an online fetch, which timed out, and then it looks in the cache. 
I however understand that this should be uniform in the code base, so I fixed it in 35238ccb5d910593b4f8d1c9ea6e6da1529eb188.

lib/viewmodels/user_comments_view_model.dart
JoachimFavre: The difference is that, when we fetch the list of comment, we want to make sure there is no other online fetch and that the UI data stay consistent everywhere (i.e. that, if the user sees 3 comments under a post, the post is indeed marked to have 3 comments). There is no such optimisation needed in this case, since deletion does not fetch a comment count.
As mentionned in the previous comment, I however removed the `setCount` for more code uniformicity, to always trust Firestore's cache.

gruvw:
I now have an error thrown to the Flutter console when deleting a comment from the user profile page:

```txt
E/flutter (17447): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Bad state: Future already completed
E/flutter (17447): #0      _AsyncCompleter.complete (dart:async/future_impl.dart:43:31)
E/flutter (17447): #1      FutureHandlerProviderElementMixin.onData (package:riverpod/src/async_notifier/base.dart:282:17)
E/flutter (17447): #2      AsyncData.map (package:riverpod/src/common.dart:345:16)
E/flutter (17447): #3      FutureHandlerProviderElementMixin.state= (package:riverpod/src/async_notifier/base.dart:205:14)
E/flutter (17447): #4      AsyncNotifierBase.state= (package:riverpod/src/async_notifier.dart:57:14)
E/flutter (17447): #5      PostCommentCountViewModel.refresh (package:proxima/viewmodels/post_comment_count_view_model.dart:23:5)
E/flutter (17447): <asynchronous suspension>
```

test/end2end/app_actions.dart
JoachimFavre: I disagree with the fact that whether we are in the last sprint changes how we test our features. We should keep testing them as well as we did before.

However, by humble curiosity, how would you make a meaningful test that tests this behaviour in an isolated? Both refreshes are done by the UI, so we did to press UI button to be able to see the view-model refresh. I could test the view-model outputs the correct value, but this is very close to testing it in the UI directly. I therefore felt like an end-to-end test was the most meaningful, hence my decision. I'm curious to hear your opinion on this.

JoachimFavre: Your comment made me think a bit more about my testing, since I do not agree with the fact that we are in the last sprint should decrease our code quality. I had decided to test my view-model using end-to-end tests because I felt I did not really have a good way of testing it more efficiently.
However, I realised I could make an integration test of this view-model without considering the UI. This is therefore what I did in 071b91a8a1dfd32d950845c5aec8d69abfa62f2c and 9afda32d11a630110ebdf9f67f0c8b10b91dc1cc, which I later fixed in 4e7a21e58ecae9836ba3adbce95b36b19586017c.

JoachimFavre:
Well, your bug is very strange. It appears the `state = const AsyncValue.loading();` I added in 810bf0977a466278e0921c81264e6ec3baea86e8 causes this issue. If I remove it, the error message disappears.
After some more digging in [riverpod's GitHub issues](https://github.com/rrousselGit/riverpod/discussions/2502), I think this is because the provider gets unmounted in the middle of its refresh, causing an issue since we try to modify the internal `state` right after. I therefore made this provider not auto-dispose in e91ddb5e432776573b49303de529196f1cb95572.
However, I'm not entirely sure neither my understanding nor my fix are correct, since I'm not sure it explains why removing the `state = const AsyncValue.loading();` would change anything, nor the exception name `Bad state: Future already completed`. @gruvw you have more experience than me in `riverpod`, I'd be very curious to hear your opinion about this.

test/end2end/app_actions.dart
```diff
@@ -322,11 +322,21 @@ class AppActions {
...
-    expect(find.text(comment), findsOneWidget);
+    final commentText = find.text(comment);
+    expect(commentText, findsOneWidget);
 
-    // Find the delete button on card
-    final deleteButton = find.byKey(ProfileInfoCard.deleteButtonCardKey);
-    expect(deleteButton, findsOneWidget);
+    // Find and tap the delete button on the card
+    final commentCard = find.ancestor(
+      of: find.text(comment),
```
Aderfish, line 330:
```suggestion
      of: commentText,
```

We can reuse the finder we declared earlier for more clarity.

Aderfish (commented):
Clean code, I just left a small suggestion. I also would like to understand more about this weird bug and its apparent fix.

Apart from that LGTM.

lib/viewmodels/user_comments_view_model.dart
gruvw: ok it makes more sense now.

lib/viewmodels/post_comment_count_view_model.dart
gruvw: Ok, I think I would have just kept `valueOrNull` instead of `value` to keep the code simpler and avoid user facing errors.

lib/viewmodels/comments_view_model.dart
gruvw: exactly, ty :+1:

test/end2end/app_actions.dart
gruvw: This wasn't about code quality but about the isolation of testing.

I think it's much better to isolate the testing of the view-model part, ty for doing it.

gruvw:
> I'd be very curious to hear your opinion about this.

I never encountered that bug and I don't really have time to dig too much into it this week.
As this provider doesn't really need to be `AutoDispose` I think we'll keep this solution for the moment.

If you think this should be investigated further, please create a new issue to the project's backlog.

gruvw (approved):
Thank you for your changes and the isolated testing. LGTM

test/end2end/app_actions.dart
JoachimFavre: I agree, this is cleaner this way! :)

Aderfish (approved):

PR #354, opened 2024-05-27
Fix errors pop ups in map page

TheTexanCodeur:
This PR fixes multiple issues linked to error pop ups in the map page. Here are the main fixs brought with this PR:

1. **Mutiple pop ups** : Error pop ups are now filtered in order to only display relevant one once (no more multiple "no location" errors).
 
2. **Errors for user posts** : Now, the map does't show an error when user checks the user posts tab with no location.

3. **Errors when location is re-enabled**: no more error pop up occur when the location is re-enabled by the user.

The last point DOES NOT WORK on the emulator. Please use a real phone to test this bug fix. 

First relevant commit: [0280f65](https://github.com/ProximaEPFL/proxima/pull/354/commits/0280f654dc1b6f1bf03c5631f614233da3178c8e)

Acceptance criteria:

- [x] The map should not show an error pop-up when the user checks the user posts on the map with location disabled
- [x] A single error pop-up message appears when the location is disabled
- [x] Error messages do not appear once the location is enabled again


lib/viewmodels/map/map_pin_view_model.dart
```diff
@@ -47,6 +47,15 @@ class MapPinViewModel extends AutoDisposeAsyncNotifier<List<MapPinDetails>> {
...
   Future<List<MapPinDetails>> _getNearbyPosts() async {
+    //checking if the location services are enabled directly here so that
+    //we throw the same exception as the challenge case
+    final locationCheck =
+        await ref.read(geolocationServiceProvider).checkLocationServices();
+
+    if (locationCheck != null) {
+      throw locationCheck;
+    }
+
```
Aderfish, line 58: Why do you do this, isn't this already done in the position stream provider ?

Aderfish (changes requested):
Good job, I see that there are less problems while manually testing on my phone. I think you should add regression tests however.

The nearby posts posts page still throws an error after toggling the location, and it never reloads correctly. I have this problem on the emulator and my real phone.

[location_toggle.webm](https://github.com/ProximaEPFL/proxima/assets/82676334/074baf3c-37af-483f-90e2-8a07e2eee114)


I think we should not show the current position marker in the my posts screen if we have location disabled, so as not to confuse the user. We could change the icon on the center button to show that we don't have location as well.

lib/viewmodels/map/map_pin_view_model.dart
TheTexanCodeur: Yes but the stream throws an exception first if i don't do it directly here. This results in different error messages when there is no location.

TheTexanCodeur:
Thanks for comment/suggestions!

I pushed a fix that should resolve the last issue.

Concerning the UI part, I think that the custom user feedback on the map when location is turned off is out of scope for this PR. The goal of this PR is to have stable error system. The features that you propose are not bug fixes in my opinion.

Don't get me wrong, I think that those suggested features are great and should be backlogged! It's just that I don't see them fit in this particular PR.

lib/viewmodels/map/map_pin_view_model.dart
Aderfish: okay that makes sense

Aderfish (approved):
Good job fixing that last problem and adding tests.

I now found a new bug where the center button no longer works after toggling location. It works again after refresh of the map page so it's not too bad. I'll add it to the backlog with my suggestions for the my posts tab. This PR indeed already accomplishes its criteria and improves the app.

PR #357, opened 2024-05-29
Navigation to map

### LINKED ISSUE #347: Navigation from post to the map view
Acceptance criteria:
- [ ] Add a button to navigate from post view to map (in nearby option) centered on the post
- [ ] Add a button to navigate from challenge list to map (in challenge option) centered on the post 
- [ ] Add a button to navigate from self post list to map (in my posts option) centered on the post

Aderfish:
Fixes #347,

This PR adds navigation from the post page, challenge page, and profile page to the map. In the post page and pop up of my posts, there is a new button, `MapActions`, that has the pin icon. On the challenge page, clicking on a challenge navigates to the map.

To do this I made a refactor of the home page navigation to use a viewmodel so that it can be accessible globaly. I have also added the possibility to pass arguments to this navigation, so that the map can be created with the correct position.

I have modified the map a bit to have an optional initial position parameter. If that parameter is null, than the map follows the user as before. Otherwise it centers on the point of interest and does not follow the user.

A lot of the file modifications are simply adding a position field to the challenge and post details classes.

I have added an extension method to transform a `GeoPoint` to a `LatLng`, so that it is easy to reuse.

image

image

Acceptance criteria:
- [x] Add a button to navigate from post view to map (in nearby option) centered on the post
- [x] Add a button to navigate from challenge list to map (in challenge option) centered on the post 
- [x] Add a button to navigate from self post list to map (in my posts option) centered on the post


lib/models/ui/challenge_details.dart
```diff
@@ -1,3 +1,4 @@
+import "package:cloud_firestore/cloud_firestore.dart";
 import "package:flutter/foundation.dart";
```
camillelnne, line 2: Is there any way to avoid using cloud firestore in the UI? Like converting to `LatLng` before passing it to the UI.
We should keep our UI and model separated.

lib/models/ui/post_details.dart
```diff
@@ -1,3 +1,4 @@
+import "package:cloud_firestore/cloud_firestore.dart";
```
camillelnne, line 1: Same comment here

lib/utils/extensions/geopoint_extensions.dart
```diff
@@ -0,0 +1,8 @@
+import "package:cloud_firestore/cloud_firestore.dart";
+import "package:google_maps_flutter/google_maps_flutter.dart";
+
+extension ToLatLng on GeoPoint {
```
camillelnne, line 4: Very smart!
I needed this at some point :)
Maybe we can add a task  in the backlog, to use it everywhere.

lib/models/ui/user_post_details.dart
```diff
@@ -1,3 +1,4 @@
+import "package:cloud_firestore/cloud_firestore.dart";
```
camillelnne, line 1: And here (tagging the code so we forget anything in case of change)

lib/views/pages/home/content/map/map_screen.dart
```diff
@@ -29,14 +36,15 @@ class MapScreen extends ConsumerWidget {
               MapSelectionOptionChips(mapInfo: value),
               const Divider(key: dividerKey),
               //TODO: change the map when clicking on a selection option
```
camillelnne, line 38:
```suggestion
```
Not your responsibility, but isn't this TODO completed?

lib/views/pages/profile/components/info_cards/profile_info_card.dart
```diff
@@ -1,3 +1,4 @@
+import "package:cloud_firestore/cloud_firestore.dart";
```
camillelnne, line 1: Same problem with the separation of UI and model here I think

lib/views/pages/profile/components/info_cards/profile_info_pop_up.dart
```diff
@@ -1,7 +1,10 @@
+import "package:cloud_firestore/cloud_firestore.dart";
```
camillelnne, line 1: ☝️ 😱

camillelnne (changes requested):
Very very cool feature!
I just have some concerns concerning the separation of view and model.

lib/utils/extensions/geopoint_extensions.dart
Aderfish: I don't think that's necessary, we can just hope that people see it and use it in the future.

lib/models/ui/challenge_details.dart
Aderfish: That makes sense. I did that in 7f2b06bc608a461b67b348f8558446793cf54333

lib/views/pages/home/content/map/map_screen.dart
Aderfish: yes I think so, a42759b3ec99b35c95f046d5d04c2e4fda332af7, (I pushed to the other map branch by accident, but we can leave it like this)

camillelnne: Thanks

camillelnne (approved):
Thanks for the changes, LGTM

PR #370, opened 2024-05-31
Cyclomatic complexity analysis

JoachimFavre:
This PR adds a simpler way to analyse cyclomatic complexity from DCM (dart code metrics), to simplify the teaching team grading task. 

To use this script, one needs to locally change the DCM parameters to make a warning for any function with cyclomatic complexity greater than 0, effectively making a warning for every function. This output is then given as parameter to the script, which outputs a, more readable, json file.

Note that we decided not to put this as part of the CI, since it would require asking for DCM to make a warning for every single function; effectively rendering it unusable for any other task.


yoannLafore (approved):
Thanks for making the script !

I have tested the script and it works perfectly.
The PR looks good to me :+1: 

Great work :tada:

